PR #289, opened 2015-06-12
use HttpContent.retrySupported if appropriate in ContentEntity

devinrsmith:


lesv:
Thanks for the submission.  I'm going to close due to staleness and lack of CLA.  Feel free to reopen if you would still like to make this change.  Ideally you'd add a test as well.

PR #290, opened 2015-06-24
Async request

hussachai:
I added new executeAsync methods to HttpRequest. The new approach is to use callback
instead of returning Future object and let users handle the result. It will be easily abused as
the get method on Future object is blocking thread. There is no point to create the new thread then block the current thread until we get the result from another thread.
The return type for executeAsync is FutureTask because it lets users cancel the submitted task.
I also added test case for those methods I created.

One more thing, the old executeAsync method that take no parameter, creates ThreadPool for every calls. That's kind of expensive call. Not sure whether it will make a memory leak or not but I'm feeling like it will. I didn't mark @depreciated on the old executeAsync and didn't fix the issue that I just mentioned. Just want to point out and let you guys decide what to do. Maybe I'm wrong.

This project is awesome because I don't like dealing with Apache HTTP Client configuration and
when I create a library relying on Apache HTTP Client. One of the user told me that Google AppEngine doesn't support it :(. Feel like google-http-java-client is a way to go.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
@@ -26,10 +26,15 @@ import com.google.api.client.util.StringUtils;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CancellationException;
+import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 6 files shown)


lesv:
Hi Thanks for the contribution.  I'm unable to accept due to the CLA not being signed.  I'm also going to close this because it's stale.  (It actually seems like a useful contribution)

PR #291, opened 2015-06-26
Catch the exception when key size mismatch

romanzhg:


ejona86:
This was synced as 7548c14

PR #292, opened 2015-06-26
avoid gzip encoding twice

danielnorberg:
gzip encoding becomes a bottle neck when e.g. publishing larger volumes of data to Google Pub/Sub.


lesv:
Thanks for the contribution.  There isn't a signed CLA and the contribution is stale so I'm going to close.

PR #295, opened 2015-07-31
Build protoc in Travis

ejona86:
This mostly fixes the Travis-CI build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,8 @@ notifications:
 
 env:
   global:
-    - PATH="/tmp/proto2.4.1/bin:$PATH"
+    - PROTO_VERSION=2.4.1
+    - PATH="/tmp/proto$PROTO_VERSION/bin:$PATH"
 
 jdk:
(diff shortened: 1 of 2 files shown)


ejona86:
@zhangkun83, would you mind reviewing this? I know it's in a strange repository :)

.travis.yml
```diff
@@ -1,6 +1,23 @@
+sudo: false
+
 language: java
+
+notifications:
+  email: false
+
+env:
+  global:
+    - PATH="/tmp/proto2.4.1/bin:$PATH"
```
zhangkun83: Why not use $PROTO_VERSION here and in the below lines where 2.4.1 is used?

ejona86: Okay. I've added PROTO_VERSION as an environment variable in the .travis.yml and use it instead of duplicating the version.

zhangkun83:
LGTM

ejona86:
Merged as 3cc6568

PR #303, opened 2015-10-26
Ignore symbolic link test failure on jdk6

ejona86:
Not much we can do; the platform simply fails to have useful behavior
for at least openjdk 6u34 and oracle jdk 6u45.

@madongfly, would you mind reviewing this (even though it is for the v1 library)?


madongfly:
@ejona86 LGTM

ejona86:
Committed as 6884507a

PR #306, opened 2015-11-03
Update httpclient dependency

dkocher:


dekobon:
Is there any traction on this PR? I would love to be able to use a new Apache HTTP Client.

ejona86:
I think the failure is a problem with a test (and not this code) that was doing I/O that ideally shouldn't. I already have a change that improves the test. Once I get it merged in I'll re-run Travis and see if it fixes the tests for this PR.

ejona86:
It looks like this code will work just fine with the old version. I may want to keep the HttpClient version at 4.0.1 and just update the code to support 4.5.1. There are some other users of this library that seem to be on old version of the library and I think they may need to be updated to support the newer version as well.

Supporting both 4.0.1 and 4.5.1 looks pretty easy and would help those other users.

dekobon:
Is there anything that I can do to help move this forward?

ejona86:
@dekobon, you could look into the failing `ApacheHttpTransportTest.testRequestsWithContent()`. I'm actually confused on how it worked before.

dekobon:
I dug pretty deeply into this. It turns out that the `com.google.api.client.testing.http.apache.MockHttpClient` is quite an insufficient implementation of a mocking layer for Apache Http Client. It attempts to allow for the internal mechanism of the `DefaultHttpClient` call stack to be executed without actually calling out to the network.

This causes problems because internal implementation has changed since version 4.1 and there are methods that are part of the underlying implementation that are being called now.

A much better mock would be an interface mock of `org.apache.http.client.HttpClient`. However, even that interface was deprecated in 4.0 of the Apache HTTP Client. That said, it is still API compatible, so it may be viable for the time being until we can get most of the consumers of this library bumped up to version 4.5. Once we have reached that point, we should rip out the deprecated calls and do things the new way.

I can submit a new PR with suggested changes. However, I'm unsure about the etiquette for this particular project.

dkocher:
Done.

ejona86:
It looks like the Travis failure is noticing the mock going away. It seems it is safe to override it. I'll have to look into that.

dekobon:
@ejona86 Could you take a look at overriding the failed travis check?

dkocher:
Any update on this?

dekobon:
Who's CLA is failing now? It shouldn't be me.

dkocher:
@dekobon It should not be me.
> It looks like you've already signed this CLA.

dkocher:
@googlebot Reply here to confirm.

dekobon:
Looking at the commit history, I don't see any committers other than dekobon and dkocher. Is the googlebot confused?

dkocher:
Confirm.

dkocher:
@ejona86 Anything required on our side?

ejona86:
@dkocher, no nothing more required from your side. The change has implications on my side and I've not had time to address them.

dkocher:
@ejona86 Ping.

gilles:
Hi,

Can you tell if this will be merged (or support for httpclient-4.5.X) be added in the foreseeable future?

There's a dependency clash with httpasyncclient-4.1.2 (requires httpclient-4.5.2). Even the lowest available in mvncentral (4.0) requires httpclient-4.3.1 which is incompatible (throws UnsupportedOperationException on `InternalHttpClient.getParams` (https://github.com/google/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java#L129)

Thanks

--Gilles

dekobon:
@gilles I override my Apache HTTP Client dependency version with 4.5.2 manually in Maven with no ill effects. However, there are some serious improvements to the codebase in the newer classes using in the 4.5.x library that would be very advantageous to use. Not to mention all of the nasty deprecation warnings that end up all over your code.

In fact, I wonder if we could just make a new transport layer that uses the latest features of Apache 4.5.x and call it something like Apache45HttpTransport. This may accommodate the legacy support needs that have been blocking @ejona86.

gilles:
If I override the dependency I run into https://github.com/google/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java#L129 throwing `UnsupportedOperationException`. But only if I create the client using `HttpClientBuilder` and then inject it (it creates an instance of `InternalHttpClient`)

If you use the default provided by google it works because it creates an instance of `DefaultHttpClient` which is deprecated.

A new `HttpTransport` would probably work. I'll try that.

dkocher:
@ejona86 Ping.

dkocher:
Any chance this would get some love?

garrettjonesgoogle:
@ejona86 it looks like you were looking at this and had context on it.

ejona86:
@garrettjonesgoogle, I've not looked at the mock changes to figure out how to fix the failures. The Mock is public API, so I'd have needed to check internally how rampant usage was, because existing usages would need to be updated.

rompic:
according to owasp dependency checks there are also some CVEs related to 4.0.1

httpclient-4.0.1.jar (cpe:/a:apache:httpclient:4.0.1, org.apache.httpcomponents:httpclient:4.0.1) : CVE-2015-5262, CVE-2014-3577, CVE-2011-1498

garrettjonesgoogle:
@vam-google could you take a look?

dkocher:
Any chance this would get (finally) some love?

saicheems:
FYI: continued at https://github.com/google/google-http-java-client/pull/377

dkocher:
Please reopen this. The crucial changes in https://github.com/google/google-http-java-client/pull/306/files#diff-49e1fb3250d450c22d6683f3aef021ddL129 are still not applied. Very tiresome contribution process here. The current dev branch will throw an exception in 
`java.lang.UnsupportedOperationException
	at org.apache.http.impl.client.InternalHttpClient.getParams(InternalHttpClient.java:211)
	at com.google.api.client.http.apache.ApacheHttpTransport.<init>(ApacheHttpTransport.java:126)`

PR #310, opened 2015-11-15
Allow unquoted field name

renfeng:
The key word, final, makes it impossible to configure the underneath Jackson implementation to allow unquoted field name for parsing. In other words, one can only hard copy the classes, and make fine tuning.


lesv:
Thanks for your contribution.  
Looks like travis / maven was having an off day. 
Since this is so stale is there any reason it's still needed?  If so, please reopen.

PR #311, opened 2015-11-18
Fixing UnsupportedOperationException on .getParams()

jarrodhroberson:
I fixed this temporarily so that I could get it to work with Apache HttpClient 4.5.1

All the `.getParams()` calls need to be migrated to `RequestConfig` instead as `.getParams()` is deprecated and now throws an `UnsupportedOperationException` I do not have time to explore refactoring to use `RequestConfig` at the moment so this small change stops the exception and lets me get on to doing what I need to do.

You can see why this is breaking based on [this question on stackoverflow](http://stackoverflow.com/questions/33761536/nt-authentication-with-googlehttpclient)


lesv:
Thanks for your contribution.  Apologies, but it appears stale so I'm going to close.

PR #312, opened 2015-11-19
Fixes for HttpClient 4.5.1 support

jarrodhroberson:
This is a cleaner solution as the other pull request I sent really didn't do anything other than eliminate the exception and it should be assumed that I have configured the HttpClient like I want.


lesv:
Thanks for the contribution.  There appears to be a merge conflict so I'm going to close.

PR #318, opened 2016-02-05
Use maven-protoc-plugin and proto 2.6.1

ejona86:
protoc 2.6.1 is available on Maven Central and maven-protoc-plugin can
automatically run it as part of the build. This prevents needing to
build protoc yourself and worrying about which protoc is in your PATH.


nmittler:
LGTM

PR #322, opened 2016-03-20
Updated to UTF-8 as the default charset

### LINKED ISSUE #300: Request charset defaults to UTF-8, Response charset defaults to ISO_8859_1
Both the request and response objects should default to the same charsets.  This causes very hard to troubleshoot bugs with repeated requests.

from `com.google.api.client.http.HttpResponse`

```
public Charset getContentCharset() {
    return mediaType == null || mediaType.getCharsetParameter() == null
        ? Charsets.ISO_8859_1 : mediaType.getCharsetParameter();
}
```

from `com.google.api.client.http.AbstractHttpContent`

```
protected final Charset getCharset() {
    return mediaType == null || mediaType.getCharsetParameter() == null
        ? Charsets.UTF_8 : mediaType.getCharsetParameter();
}
```

Ideally both will use `Charsets.UTF_8`, as `Charsets.ISO_8859_1` only supports a subset.

ameya-pandilwar:
Fixes #300

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -518,12 +518,12 @@ public final class HttpResponse {
 
   /**
    * Returns the {@link Charset} specified in the Content-Type of this response or the
-   * {@code "ISO-8859-1"} charset as a default.
+   * {@code "UTF-8"} charset as a default.
    *
    * @since 1.10
    * */
   public Charset getContentCharset() {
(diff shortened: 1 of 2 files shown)


elharo:
IMHO the fix needs to go the other way (specify ISO-8859-1 when there's no header) for spec conformance. 

We really should have a test for this, one way or the other.

lesv:
I'm going to leave this open, but not merge to save @elharo 's coment / review.  My thought was about limiting changes to existing behavior.

chingor13:
Closing as this is not the desired behavior.

PR #333, opened 2016-09-08
Swallow exception from close call when there is an underlying IOException

chajath:
There have been numerous reports of users of the client library seeing IO exceptions bearing the message "insufficient data written" (google/google-api-java-client#781 google/google-api-java-client#782 GoogleCloudPlatform/google-cloud-java#1018)

I believe these error messages are misleading, and are secondary effect of attempting to close the stream when the underlying IO operation was actually failing (see [underlying close implementation](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L3187)) 

Therefore I propose we handle the case when there has been an underlying IO exception error, we still attempt to close the stream but let the original IO exception propagate nonetheless.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
@@ -75,12 +75,10 @@ final class NetHttpRequest extends LowLevelHttpRequest {
           connection.setChunkedStreamingMode(0);
         }
         OutputStream out = connection.getOutputStream();
-        IOException thrown = null;
+        boolean threw = true;
         try {
           getStreamingContent().writeTo(out);
-        } catch(IOException exception) {
-          thrown = exception;
-          throw exception;
+          threw = false;
         } finally {
           try {
             out.close();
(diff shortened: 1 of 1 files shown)


vikkyrk:
@chajath any updates on this? We are seeing some Dataflow customers facing issues where they cannot debug the underlying exception thrown? Do you know when will this be released?

chajath:
I haven't heard anything back yet. I will follow up with the team if I don't hear back soon.

james-woods:
As a Dataflow customer we would really love to see some progress on this (:

chajath:
There's also #334 that I've opened that should fix the build.

vikkyrk:
@chajath @ejona86 any updates? We (Datastore/Dataflow teams) are waiting for this change.

JoshuaFox:
Thanks, looking forward to the fix. I'm the Dataflow user. This fix will help us debug what is happening in the my Pipeline --- at the moment, there is no info on what is making it fail.

PR #347, opened 2017-02-22
NetHttpRequest should only set Content-Length if it's not set.

monktastic:
See https://github.com/google/google-http-java-client/issues/346.

The reason this bug isn't more common is probably that nobody sets sun.net.http.allowRestrictedHeaders=true. If they do, and they set Content-Length, then NetHttpClient will duplicate it.


monktastic:
CLA signed.

PR #349, opened 2017-03-10
Fix url path escaping

david-barrett:


lesv:
Thank you for your behavior.  I'm going to close this as I'm concerned about changing existing behavior.  If you still need this could you mention why and reopen.

PR #352, opened 2017-03-23
Update to Guava 20.0

ronshapiro:
guava-jdk5 17.0 is very old: 3 years, which is beyond the supported lifetime of Guava. Additionally, guava on jdk5 is no longer supported since Java 5 has been EOL for a while.

This library depending on a very old version of guava was first reported here: https://github.com/google/dagger/issues/651

I don't know what to do with the files in `google-http-client-assembly/dependencies/` but those look to be out of date anyway.

Note that I didn't update to Guava 21, since that version requires Java 8.

### DIFF
--- a/google-http-client-appengine/pom.xml
+++ b/google-http-client-appengine/pom.xml
@@ -80,7 +80,7 @@
     </dependency>
     <dependency>
       <groupId>com.google.guava</groupId>
-      <artifactId>guava-jdk5</artifactId>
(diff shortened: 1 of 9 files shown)


garrettjonesgoogle:
I don't think we should go past 19.0 for now. See https://github.com/grpc/grpc-java/pull/2751

ronshapiro:
Hm. While that would probably have other benefits, that wouldn't solve the referenced issue.

PR #377, opened 2017-10-10
Bump version of Apache httpclient.

mattwhisenhunt:
@garrettjonesgoogle

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
@@ -123,6 +123,12 @@ public final class ApacheHttpTransport extends HttpTransport {
    */
   public ApacheHttpTransport(HttpClient httpClient) {
     this.httpClient = httpClient;
+    HttpParams params = httpClient.getParams();
+    if (params == null) {
+      params = newDefaultHttpClient().getParams();
+    }
+    HttpProtocolParams.setVersion(params, HttpVersion.HTTP_1_1);
+    params.setBooleanParameter(ClientPNames.HANDLE_REDIRECTS, false);
   }
 
   /**
@@ -161,8 +167,6 @@ public final class ApacheHttpTransport extends HttpTransport {
     HttpConnectionParams.setSocketBufferSize(params, 8192);
     ConnManagerParams.setMaxTotalConnections(params, 200);
     ConnManagerParams.setMaxConnectionsPerRoute(params, new ConnPerRouteBean(20));
-    HttpProtocolParams.setVersion(params, HttpVersion.HTTP_1_1);
-    params.setBooleanParameter(ClientPNames.HANDLE_REDIRECTS, false);
     return params;
   }
 
--- a/google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
@@ -123,9 +123,6 @@ public final class ApacheHttpTransport extends HttpTransport {
    */
   public ApacheHttpTransport(HttpClient httpClient) {
     this.httpClient = httpClient;
-    HttpParams params = httpClient.getParams();
-    HttpProtocolParams.setVersion(params, HttpVersion.HTTP_1_1);
-    params.setBooleanParameter(ClientPNames.HANDLE_REDIRECTS, false);
   }
 
   /**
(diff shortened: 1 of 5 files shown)


google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
```diff
@@ -123,9 +123,6 @@ public ApacheHttpTransport() {
    */
   public ApacheHttpTransport(HttpClient httpClient) {
     this.httpClient = httpClient;
-    HttpParams params = httpClient.getParams();
-    HttpProtocolParams.setVersion(params, HttpVersion.HTTP_1_1);
-    params.setBooleanParameter(ClientPNames.HANDLE_REDIRECTS, false);
```
garrettjonesgoogle: It looks like this code has no effect - so it was broken before your change?

google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
```diff
@@ -278,7 +277,7 @@ public HttpClient getHttpClient() {
      * </p>
      *
      * <pre>
-       setProxy(new HttpHost("127.0.0.1", 8080))
+     setProxy(new HttpHost("127.0.0.1", 8080))
```
garrettjonesgoogle: Within a `<pre>` block, you can still include the stars (`*`) at the beginning of the line (for a more consistent look), and they are stripped out before generating the javadoc output.

google-http-client/src/test/java/com/google/api/client/http/apache/ApacheHttpTransportTest.java
```diff
@@ -43,17 +47,16 @@ public void testApacheHttpTransport() {
...
-  }
-
   public void testNewDefaultHttpClient() {
     checkDefaultHttpClient(ApacheHttpTransport.newDefaultHttpClient());
   }
 
   public void testRequestsWithContent() throws Exception {
-    ApacheHttpTransport transport = new ApacheHttpTransport(new MockHttpClient());
+    HttpClient mockClient = mock(HttpClient.class);
+    org.apache.http.HttpResponse mockResponse = mock(org.apache.http.HttpResponse.class);
```
garrettjonesgoogle: Why does `HttpResponse` need to be fully qualified?

google-http-client/src/main/java/com/google/api/client/testing/http/apache/MockHttpClient.java
```diff
@@ -50,13 +50,14 @@
  * @since 1.14
  * @author Yaniv Inbar
  */
-@Beta
+@Deprecated
```
garrettjonesgoogle: Let's give more details about the deprecation - e.g. can it not be used any more with this library because of the upgrade of the Apache http client version?

google-http-client/src/test/java/com/google/api/client/http/apache/ApacheHttpTransportTest.java
mattwhisenhunt: No need. :) Thanks.

google-http-client/src/main/java/com/google/api/client/testing/http/apache/MockHttpClient.java
mattwhisenhunt: clirr (a Maven plugin) kept failing other PRs because this file was moved all together. I was going to deprecate it in this version and remove it completely in the future.

mattwhisenhunt:
@garrettjonesgoogle PTAL

google-http-client/src/test/java/com/google/api/client/http/apache/ApacheHttpTransportTest.java
```diff
@@ -43,17 +48,16 @@ public void testApacheHttpTransport() {
     checkHttpClient(httpClient);
   }
 
-  public void testApacheHttpTransportWithParam() {
-    ApacheHttpTransport transport = new ApacheHttpTransport(new DefaultHttpClient());
-    checkHttpClient(transport.getHttpClient());
-  }
```
garrettjonesgoogle: Does this test need to stay deleted? Are we losing coverage of `transport.getHttpClient()`?

mattwhisenhunt: No. Yes.

Thank you. A new commit just fixed that.

google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
```diff
@@ -124,6 +124,9 @@ public ApacheHttpTransport() {
   public ApacheHttpTransport(HttpClient httpClient) {
     this.httpClient = httpClient;
     HttpParams params = httpClient.getParams();
```
dkocher: Deprecated usage that will throw an `UnsupportedOperationException`.

PR #382, opened 2017-11-10
Log exception only when logging is enabled

mahiuchun:


google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -999,7 +999,9 @@ public HttpResponse execute() throws IOException {
         }
         // Save the exception in case the retries do not work and we need to re-throw it later.
         executeException = e;
-        logger.log(Level.WARNING, "exception thrown while executing request", e);
+        if (loggingEnabled) {
```
saicheems: s/loggingEnabled/loggable

mahiuchun: Done

PR #385, opened 2017-12-06
Fixing ArrayMap hashCode to work with null values

KotaiVictor:
This should fix: https://github.com/google/google-http-java-client/issues/384

Just checking if the values are not null before calling the hash code method on them.


KotaiVictor:
@mattwhisenhunt I don't have rights to merge this in. Could you do it if everything's in order?

jadekler:
We are in the process of releasing, and fixing the release process. This process will take several weeks unfortunately; we'll post updates upon completion.

PR #393, opened 2018-01-28
Add OpenCensus tracing instrumentation in google-http-client module.

HailongWen:
This PR adds [OpenCensus-Java](https://github.com/census-instrumentation/opencensus-java) instrumentation in to monitor and record events when executing `HttpExecute#execute()`.

The instrumentation starts a `Span` when the method begins, and is ended before the method exits. The status of the span will be determined according to the status of the response. This `Span` is stored into local span store, and can be optionally exported if user explicitly configure an exporter (e.g. [stackdriver exporter](https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/trace/stackdriver)).

`Span`s can have children. In this case, the generated `Span` in the instrumentation will be parent of all the `Span`s generated by server-side procedure. To link the children `Span` with their parent, the context information should be propagated to server side. This is done by using [OpenCensus HTTP Util](https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/http_util) to inject the information into the header in the request.

Added `OpenCensusUtils` provides helper methods to perform context propagation, and setting end status of span.

The added dependency of `opencensus-api` contains basic no-op implementation. It will be replaced by real implementation if user explicitly adds dependency of `opencensus-impl` or `opencensus-impl-lite` (for android and gae) in the classpath.

This PR does not contain integration test when `opencensus-impl` is provided. It can be added with the help of maintainers.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
@@ -18,10 +18,15 @@ import com.google.api.client.util.Beta;
 import com.google.api.client.util.IOUtils;
 import com.google.api.client.util.LoggingStreamingContent;
 import com.google.api.client.util.ObjectParser;
+import com.google.api.client.util.OpenCensusUtils;
 import com.google.api.client.util.Preconditions;
 import com.google.api.client.util.Sleeper;
 import com.google.api.client.util.StreamingContent;
 import com.google.api.client.util.StringUtils;
+
+import io.opencensus.common.Scope;
+import io.opencensus.trace.Span;
+import io.opencensus.trace.Tracer;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -209,6 +214,12 @@ public final class HttpRequest {
 
   /** Sleeper. */
   private Sleeper sleeper = Sleeper.DEFAULT;
+
+  /** OpenCensus tracing component. */
+  private Tracer tracer = OpenCensusUtils.getTracer();
+
+  /** Prefix for tracing span name. */
+  private static final String traceSpanNamePrefix = "Sent." + HttpRequest.class.getName() + ".";
 
   /**
    * @param transport HTTP transport
@@ -854,7 +865,9 @@ public final class HttpRequest {
     Preconditions.checkNotNull(requestMethod);
     Preconditions.checkNotNull(url);
 
+    Span span = tracer.spanBuilder(traceSpanNamePrefix + "execute").startSpan();
     do {
+      span.addAnnotation("retry #" + numRetries);
       // Cleanup any unneeded response from a previous iteration
       if (response != null) {
         response.ignore();
@@ -898,6 +911,8 @@ public final class HttpRequest {
           headers.setUserAgent(originalUserAgent + " " + USER_AGENT_SUFFIX);
         }
       }
+      OpenCensusUtils.propagateTracingContext(headers);
+
       // headers
       HttpHeaders.serializeHeaders(headers, logbuf, curlbuf, logger, lowLevelHttpRequest);
       if (!suppressUserAgentSuffix) {
@@ -977,6 +992,8 @@ public final class HttpRequest {
 
       // execute
       lowLevelHttpRequest.setTimeout(connectTimeout, readTimeout);
+      // switch tracing scope to current span
+      Scope ws = tracer.withSpan(span);
       try {
         LowLevelHttpResponse lowLevelHttpResponse = lowLevelHttpRequest.execute();
         // Flag used to indicate if an exception is thrown before the response is constructed.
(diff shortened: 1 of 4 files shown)


HailongWen:
Hi @mattwhisenhunt , would you please kindly help on this PR?

google-http-client/src/main/java/com/google/api/client/util/OpenCensusUtils.java
```diff
@@ -0,0 +1,157 @@
...
+   * {@link TextFormat.Setter} for {@link activeTextFormat}.
+   */
+  @Nullable
+  static TextFormat.Setter propagationTextFormatSetter = null;
+
+  /**
+   * Sets the {@link TextFormat} used in context propagation.
+   * @param textFormat the text format.
+   */
+  public static void setPropagationTextFormat(@Nullable TextFormat textFormat) {
```
bogdandrutu: Why is this public? Should be package protected and probably with an annotation that is for testing only.

google-http-client/src/main/java/com/google/api/client/util/OpenCensusUtils.java
```diff
@@ -0,0 +1,157 @@
...
+    propagationTextFormat = textFormat;
+  }
+
+  /**
+   * Sets the {@link TextFormat.Setter} used in context propagation.
+   * @param textFormatSetter the {@code TextFormat.Setter} for the text format.
+   */
+  public static void setPropagationTextFormatSetter(@Nullable TextFormat.Setter textFormatSetter) {
```
bogdandrutu: same comment.

google-http-client/src/main/java/com/google/api/client/util/OpenCensusUtils.java
```diff
@@ -0,0 +1,157 @@
...
+   * {@link TextFormat} used in tracing context propagation.
+   */
+  @Nullable
+  static TextFormat propagationTextFormat = null;
+
+  /**
+   * {@link TextFormat.Setter} for {@link activeTextFormat}.
+   */
+  @Nullable
+  static TextFormat.Setter propagationTextFormatSetter = null;
```
bogdandrutu: Because you allow setters you must comment why this is not volatile? If one thread sets this it is not guaranteed that the other threads will see the newest value.

This will work because you actually initialize this in the static initialization code.

bogdandrutu: Also why this and propagationTextFormat are not private (if for testing please comment or use the annotation)? Same comment for tracer.

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -854,7 +865,9 @@ public HttpResponse execute() throws IOException {
     Preconditions.checkNotNull(requestMethod);
     Preconditions.checkNotNull(url);
 
+    Span span = tracer.spanBuilder(traceSpanNamePrefix + "execute").startSpan();
```
bogdandrutu: traceSpanNamePrefix + "execute" this is a constant string, you should use a constant for this to avoid unnecessary overhead.

google-http-client/src/main/java/com/google/api/client/util/OpenCensusUtils.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  /**
+   * Propagate information of current tracing context. This information will be injected into HTTP
+   * header.
+   */
+  public static void propagateTracingContext(HttpHeaders headers) {
+    Preconditions.checkNotNull(headers);
+    if (propagationTextFormat != null && propagationTextFormatSetter != null) {
+      Span span = tracer.getCurrentSpan();
```
bogdandrutu: Better to pass the Span as argument.

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -1057,6 +1076,7 @@ public HttpResponse execute() throws IOException {
         }
       }
     } while (retryRequest);
+    span.end(OpenCensusUtils.getEndSpanOptions(response == null ? null : response.getStatusCode()));
```
bogdandrutu: Before ending the span is good to record a MessageEvent with the received message size.

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -977,6 +992,8 @@ public HttpResponse execute() throws IOException {
 
       // execute
       lowLevelHttpRequest.setTimeout(connectTimeout, readTimeout);
+      // switch tracing scope to current span
+      Scope ws = tracer.withSpan(span);
```
bogdandrutu: Before sending the request probably want to record a MessageEvent (network event) with the size of the message sent.

PR #394, opened 2018-01-31
Enhance OpenCensus tracing instrumentation.

HailongWen:
* Add method and instrumentation to record message events with content-length when sending/receiving
request/response.

* Refine API, add `volatile` modifiers to internal attributes.

* Use `registerSpanNamesForCollection` to replace `setSampleToLocalSpanStore`.

* Add get/set method for recordEvent so users can controll whether to expose collected spans.

* Add more tests and rewrite/fix test cases with JUnit 3 style.

* Fix/refine javadoc and messages.

I have done some basic integration test locally and fix some defects. Codes can be found [here](https://github.com/HailongWen/opencensus-java/blob/google-http-java-client-interop-testing/examples/src/test/java/io/opencensus/examples/trace/GoogleHttpJavaClientInteropTesting.java)

Please @mattwhisenhunt kindly help on merging this PR after @bogdandrutu 's review.
There might be 1 more PR on stats collecting.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
@@ -25,11 +25,14 @@ import com.google.api.client.util.StreamingContent;
 import com.google.api.client.util.StringUtils;
 
 import io.opencensus.common.Scope;
+import io.opencensus.trace.Annotation;
+import io.opencensus.trace.AttributeValue;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Collections;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
@@ -866,8 +869,15 @@ public final class HttpRequest {
         .spanBuilder(OpenCensusUtils.SPAN_NAME_HTTP_REQUEST_EXECUTE)
         .setRecordEvents(OpenCensusUtils.isRecordEvent())
         .startSpan();
+    long idGenerator = 0L;
+
     do {
-      span.addAnnotation("retry #" + (numRetries - retriesRemaining));
+      span.addAnnotation(
+          Annotation.fromDescriptionAndAttributes(
+              "retry",
+              Collections.<String, AttributeValue>singletonMap(
+                  "number of retry",
+                  AttributeValue.longAttributeValue(numRetries - retriesRemaining))));
       // Cleanup any unneeded response from a previous iteration
       if (response != null) {
         response.ignore();
@@ -911,7 +921,7 @@ public final class HttpRequest {
           headers.setUserAgent(originalUserAgent + " " + USER_AGENT_SUFFIX);
         }
       }
-      OpenCensusUtils.propagateTracingContext(span, headers);
+      OpenCensusUtils.propagateTracingContext(span.getContext(), headers);
 
       // headers
       HttpHeaders.serializeHeaders(headers, logbuf, curlbuf, logger, lowLevelHttpRequest);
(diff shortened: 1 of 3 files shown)


google-http-client/src/main/java/com/google/api/client/util/OpenCensusUtils.java
```diff
@@ -42,26 +48,49 @@
...
    * OpenCensus tracing component.
    * When no OpenCensus implementation is provided, it will return a no-op tracer.
    */
-  static Tracer tracer = Tracing.getTracer();
+  private static Tracer tracer = Tracing.getTracer();
+
+  /**
+   * Sequence id generator for message event.
+   */
+  private static AtomicLong idGenerator = new AtomicLong();
```
bogdandrutu: This should be per request not global. In an ideal world this should match with the message event on the server so better to start with 0 for each request.

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -865,9 +862,12 @@ public HttpResponse execute() throws IOException {
...
     Preconditions.checkNotNull(url);
 
-    Span span = tracer.spanBuilder(traceSpanNamePrefix + "execute").startSpan();
+    Span span = tracer
+        .spanBuilder(OpenCensusUtils.SPAN_NAME_HTTP_REQUEST_EXECUTE)
+        .setRecordEvents(OpenCensusUtils.isRecordEvent())
+        .startSpan();
     do {
-      span.addAnnotation("retry #" + numRetries);
+      span.addAnnotation("retry #" + (numRetries - retriesRemaining));
```
bogdandrutu: probably for the number of retries you can use an annotation with an attribute, it has the type int.

google-http-client/src/test/java/com/google/api/client/util/OpenCensusUtilsTest.java
```diff
@@ -36,9 +40,11 @@
  */
 public class OpenCensusUtilsTest extends TestCase {
 
-  @Rule public ExpectedException thrown = ExpectedException.none();
   TextFormat mockTextFormat;
   TextFormat.Setter mockTextFormatSetter;
+  TextFormat originTextFormat;
+  TextFormat.Setter originTextFormatSetter;
+  Span mockSpan;
```
bogdandrutu: Why do you not use real mock objects? Like use mockito?

HailongWen: It seems that most of the test mocks are written manually so I just followed that convention, Besides all the tests are written in JUnit3 style.

But yeah I think that is a good idea. I'll rewrite that.

HailongWen:
@bogdandrutu PTAL.

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -865,9 +865,19 @@ public HttpResponse execute() throws IOException {
...
+        .startSpan();
+    long idGenerator = 0L;
+
     do {
-      span.addAnnotation("retry #" + numRetries);
+      span.addAnnotation(
+          Annotation.fromDescriptionAndAttributes(
+              "retry",
+              Collections.<String, AttributeValue>singletonMap(
+                  "number of retry",
```
bogdandrutu: just "number"

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -994,8 +1004,14 @@ public HttpResponse execute() throws IOException {
       lowLevelHttpRequest.setTimeout(connectTimeout, readTimeout);
       // switch tracing scope to current span
       Scope ws = tracer.withSpan(span);
+      OpenCensusUtils.recordSentMessageEvent(
+          span, idGenerator++, lowLevelHttpRequest.getContentLength());
```
bogdandrutu: you need an idGenerator for sent and one for receive.

HailongWen:
Thanks for the review!

HailongWen:
Thanks Matt!

PR #404, opened 2018-04-15
Make json serialization treat maps as maps, always.

mikekap:
Even if they are iterable, like clojure maps. See https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1890 . See https://github.com/clojure/clojure/blob/master/src/jvm/clojure/lang/APersistentMap.java#L16 .


jadekler:
Hello - thanks for contributing. Just a quick update that this repository is currently frozen whilst we fix some backend stuff related to the release process, and submit a release. It should be unfrozen in a few weeks.

More generally, this repository is in maintenance mode; we may not be able to get to this PR for some time.

mikekap:
@chingor13 I'm not 100% sure where to put the tests - it doesn't look like there are any generator tests currently - only parser tests. From my perspective, I'm using https://googlecloudplatform.github.io/google-cloud-java/google-cloud-clients/apidocs/com/google/cloud/bigquery/InsertAllRequest.Builder.html#addRow-java.lang.String-java.util.Map- - which takes an arbitrary map. This currently fails spectacularly for Clojure maps, but generally an implementation of `Map` that looks like this will fail:

```java
import java.util.HashMap;
import java.util.Iterator;
import java.util.Map;
import java.util.Spliterator;
import java.util.function.Consumer;

public class IterableMap extends HashMap<String, String> implements Iterable<Map.Entry<String, String>> {
    @Override
    public Iterator<Entry<String, String>> iterator() {
        return entrySet().iterator();
    }

    @Override
    public void forEach(Consumer<? super Entry<String, String>> action) {
        entrySet().forEach(action);
    }

    @Override
    public Spliterator<Entry<String, String>> spliterator() {
        return entrySet().spliterator();
    }
}
```

chingor13:
Bleh, GitHub did not like the merge from master.

chingor13:
@mikekap Sorry, I messed up your branch trying to merge from master (to run CI). I created #550 with the added tests and cherry-picked your commit into there.

PR #406, opened 2018-04-18
internal: resolve circular dependency

jadekler:
Resolves a circular dependency:

`utils.OpenCensusUtils` depends on `http.HttpHeaders`, `http.HttpHeaders` depends on `utils.*`.

The change is to move `utils.OpenCensusUtils` to `http.OpenCensusUtils`, which is not the greatest solution but should work.

Maven doesn't care but google's internal build tooling does care.


jadekler:
cc @neozwu

PR #409, opened 2018-04-24
Merge from internal google

jadekler:
DO NOT MERGE

**Note**: this PR includes a change to enforce Java 1.7 (previously 1.5), since several parts of this PR do not build in 1.5.

Moe writing branch from 79e973276b27cb4a873097b7025fcebfff1d00a4

### DIFF
--- a/google-http-client-jdo/pom.xml
+++ b/google-http-client-jdo/pom.xml
@@ -54,7 +54,7 @@
       </plugin>
       <plugin>
         <groupId>org.datanucleus</groupId>
-        <artifactId>maven-datanucleus-plugin</artifactId>
+        <artifactId>datanucleus-maven-plugin</artifactId>
         <configuration>
           <verbose>true</verbose>
         </configuration>
@@ -75,6 +75,18 @@
       <artifactId>google-http-client</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.datanucleus</groupId>
+      <artifactId>datanucleus-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.datanucleus</groupId>
+      <artifactId>datanucleus-api-jdo</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.datanucleus</groupId>
+      <artifactId>javax.jdo</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.http-client</groupId>
       <artifactId>google-http-client-test</artifactId>
       <scope>test</scope>
@@ -82,20 +94,6 @@
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>javax.jdo</groupId>
-      <artifactId>jdo2-api</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.datanucleus</groupId>
-      <artifactId>datanucleus-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.datanucleus</groupId>
-      <artifactId>datanucleus-api-jdo</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
--- a/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
@@ -74,26 +74,11 @@ final class NetHttpRequest extends LowLevelHttpRequest {
         } else {
           connection.setChunkedStreamingMode(0);
         }
-        OutputStream out = connection.getOutputStream();
-        boolean threw = true;
-        try {
+
+        try (OutputStream out = connection.getOutputStream()) {
           getStreamingContent().writeTo(out);
-          threw = false;
-        } finally {
-          try {
-            out.close();
-          } catch (IOException exception) {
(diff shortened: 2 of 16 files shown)


pom.xml
```diff
@@ -620,7 +620,7 @@
     <project.jackson-core-asl.version>1.9.11</project.jackson-core-asl.version>
     <project.jackson-core2.version>2.9.2</project.jackson-core2.version>
     <project.protobuf-java.version>2.6.1</project.protobuf-java.version>
-    <project.guava.version>19.0</project.guava.version>
+    <project.guava.version>24.1-jre</project.guava.version>
```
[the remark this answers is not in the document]
jadekler: Shoot! I double checked and thankfully it's 20 [1], and 20 supports `throwIfUnchecked`. 

1: Don't know how legit https://stackoverflow.com/questions/36045121/which-version-of-guava-is-compatible-with-java-1-7 is... let me know if you know otherwise!

mattwhisenhunt:
Its legit: https://github.com/google/guava/wiki/Compatibility

jadekler:
@lesv @ejona86 I ran into build problems with 1.5. I upgraded to 1.7, which fixes the build problems, but I'm not sure whether 1.5 is a requirement. PTAL and let me know.

google-http-client-jdo/pom.xml
```diff
@@ -75,27 +75,25 @@
       <artifactId>google-http-client</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.google.http-client</groupId>
-      <artifactId>google-http-client-test</artifactId>
-      <scope>test</scope>
+      <groupId>org.datanucleus</groupId>
```
ejona86: How did you chose this group id?

google-http-client-jdo/pom.xml
```diff
@@ -75,27 +75,25 @@
...
+      <artifactId>datanucleus-core</artifactId>
```
ejona86: It looks like previously this code depended against just the jdo API. Is this core just API, or is it implementation?

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
```diff
@@ -74,31 +74,17 @@ public LowLevelHttpResponse execute() throws IOException {
         } else {
           connection.setChunkedStreamingMode(0);
         }
-        OutputStream out = connection.getOutputStream();
-        boolean threw = true;
-        try {
+
+        try (OutputStream out = connection.getOutputStream()) {
```
ejona86: Yeah... We don't do that in this code base, because of the Java version restriction.

google-http-client/src/main/java/com/google/api/client/util/Base64.java
```diff
@@ -14,6 +14,8 @@
 
 package com.google.api.client.util;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
```
ejona86: And we don't do this either, because of version restriction.

pom.xml
```diff
@@ -249,25 +249,26 @@
         <artifactId>mockito-all</artifactId>
         <version>1.9.0</version>
       </dependency>
-      <dependency>
-        <groupId>javax.jdo</groupId>
```
ejona86: Why were these jdo changes made? These Maven files wouldn't really have been modified internally and I don't see any Java code modifications that would trigger this change.

ejona86:
@jadekler, yes, swapping from 1.5 to 1.7 shouldn't be done lightly. There was already strong interest in dropping 1.5, since _surely_ there aren't many using it. But 1.6 should be handled a bit more warily.

What parts conflict? Supporting 1.5 is the main reason we review all CLs for this code, as the general assumption is 1.8+ features are available and we don't want those to go in internally because we couldn't bring them external.

ejona86:
I see there's a fix here for "Fix datanucleus build errors". If Travis or such is broken on master, that commit should be done as a separate PR. Right now it looks like it's part of the "merge from internal".

jadekler:
> @jadekler, yes, swapping from 1.5 to 1.7 shouldn't be done lightly. There was already strong interest in dropping 1.5, since surely there aren't many using it. But 1.6 should be handled a bit more warily.

Yikes, ok - wasn't aware we were supporting versions this far back. I'm not sure that we can two-way sync successfully _and_ support versions of java that far back. Many of the java warnings we get internally are not compatible with java 5 (and I would venture a guess that several of them are not compatible with java 6).

> I see there's a fix here for "Fix datanucleus build errors". If Travis or such is broken on master, that commit should be done as a separate PR. Right now it looks like it's part of the "merge from internal".

Master is green - "fix datanucleus build errors" was to resolve the travis build errors that "checkstyle plugin upgrade and fixes" cause.

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
jadekler: Ok, good context. Thank you.

ejona86:
> Many of the java warnings we get internally are not compatible with java 5

Right. We just leave those. If a Java style reviewer complains, the response "we support Java 5" has generally been adequate. (Maybe followed by some o.O looks.)

> "fix datanucleus build errors" was to resolve the travis build errors that "checkstyle plugin upgrade and fixes" cause.

Why was checkstyle upgraded in this PR?

jadekler:
Abandoned, opened new PR that has none of the linting fixes https://github.com/google/google-http-java-client/pull/410

PR #412, opened 2018-05-02
google-http-client : For controlling whether to pass authorization header in redirected request or not.

raghuhit:
- Apache http client and many other clients, by default allow authorization headers to be passed to redirected request.

- Here I have added one parameter , "passAuthHeaderInRedirects" is added for enabling and disabling of passing authorization header in redirections.


raghuhit:
I signed it

PR #414, opened 2018-05-20
supress MustBeClosed

jadekler:
This code does not compile internally.

Internal error: 

```
error: [MustBeClosedChecker] The result of this method must be closed.
      Scope ws = tracer.withSpan(span);
```

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
@@ -1003,6 +1003,7 @@ public final class HttpRequest {
       // execute
       lowLevelHttpRequest.setTimeout(connectTimeout, readTimeout);
       // switch tracing scope to current span
+      @SuppressWarnings("MustBeClosedChecker")
       Scope ws = tracer.withSpan(span);
       OpenCensusUtils.recordSentMessageEvent(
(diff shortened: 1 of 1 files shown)


jadekler:
cc @HailongWen

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -1003,7 +1003,6 @@ public HttpResponse execute() throws IOException {
       // execute
       lowLevelHttpRequest.setTimeout(connectTimeout, readTimeout);
       // switch tracing scope to current span
```
mattwhisenhunt: Remove this comment too.

jadekler: Done

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -1002,8 +1002,6 @@ public HttpResponse execute() throws IOException {
 
       // execute
       lowLevelHttpRequest.setTimeout(connectTimeout, readTimeout);
-      // switch tracing scope to current span
-      Scope ws = tracer.withSpan(span);
```
HailongWen: This line switches the context to the newly created span.
We always create a new span in case user does not have one.

PR #417, opened 2018-06-21
base64: strip carriage returns and newlines

jadekler:
The guava [1] doc mentions that carriage returns and newlines are not
permitted, and recommend using .withSeparator. We do not know whether
the input string will contain a carriage return, or a newline, or both,
so instead we'll just strip them each individually.

This fixes an internal google bug that was introduced in as part of
the move from Apache Commons Codec to Guava in
ef9f9142be5dacdd19a3a16f37000eca478d2d51.

1: https://google.github.io/guava/releases/19.0/api/docs/com/google/common/io/BaseEncoding.html#withSeparator(java.lang.String,%20int)

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/Base64.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/Base64.java
@@ -105,6 +105,9 @@ public class Base64 {
       return null;
     }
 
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
@jadekler lets start adding @google/yoshi-java to these reviews :)

PR #418, opened 2018-06-25
Rollback base64 and opencensus

jadekler:
- Adding opencensus means depending on instrumentation, which depends on grpc. For certain android projects, this makes dex files too large, causing their compilation to fail. The solution is not immediately obvious to me, and probably requires a lot of communication with the failing teams. This is time that we can not spend at the moment.

- Opencensus causes some overly-tightly-coupled tests to fail. This is trivial to solve but takes time.

- The Base64 encoding change causes several tests internally to fail. This is non-trivial to solve and may take some time and several iterations.


jadekler:
@HailongWen PTAL

HailongWen:
I am fine with the rollback of OpenCensus for now.
It must be noted that the dependency issue could be related to internal code change and need some time to resolve. And when it is resolved, we should work on another PR to add it back.

PR #420, opened 2018-07-10
Fixed calls to toLowerCase() to be toLowerCase(Locale.US) instead.

jadekler:
The no-arg method uses Locale.getDefault() internally. On an Android device, this uses the currently active Locale. If the currently active locale happens to be Turkish, then uppercase "I" becomes a non-ASCII unicode character 'LATIN SMALL LETTER DOTLESS I'.

This specifically manifested as the header X-Goog-Encode-Response-If-Executable being rejected by the okhttp library because it contained the non-ASCII character.

NOTE: this commit was originally written by @stuartfehr


jadekler:
(I'll hold off on merging this in until we have finished the release that is in progress)

PR #422, opened 2018-07-12
generated commit: update to 1.25.0-SNAPSHOT

jadekler:


jadekler:
Regenerating with synced dependency versions.

PR #425, opened 2018-07-16
docs: document process for updating pom.xml versions

jadekler:
I intend on making the same change to oauth and api, but figured I'd just send out one CL for now in order to codereview phrasing in only one place (I will use this as a template for the other repos).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -548,6 +548,15 @@
     </plugins>
   </build>
   <properties>
+    <!--
+      NOTE: if you make a change to these versions, you MUST make the same change to:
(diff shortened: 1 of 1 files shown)


chingor13:
It might also be worthwhile to add a `RELEASING.md` similar to [google-cloud-java](https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/RELEASING.md), [google-auth-library-java](https://github.com/google/google-auth-library-java/blob/master/RELEASE.md)

jadekler:
I'd really like that, but I don't think we can, since it is 99% internal work.

That said, I am building a `RELEASING.md` internally.

PR #434, opened 2018-07-25
remove <keep>com.google.api.client.**, fixing packaging

jadekler:
Before this commit, packaging is broken: the google oauth and api libraries depend on this library repackaging guava in a specific way. Parts of the jarjar configuration (above the keep declaration) do just that. However, for some reason, not all of the guava classes seem to be included: for instance, oauth will complain that the following class is missing: com/google/api/client/repackaged/com/google/common/base/Platform.

What gives? guava-20 certainly contains that class. Even more strange, there _are_ things in that directory, just not _all_ of the classes expected. So, _some_ guava things are obviously being repackaged.

Well, the inclusion of the keep rule removed in this commit somehow seems to either:
- Repackage a very old version of guava that does not have all the classes that newer code expects
- Remove some classes

I suspect the former. Either way, this change makes all three libraries `mvn clean install` without a problem, and does not seem to have a negative impact (e.g., the existence of com.google.api.client classes in the jar does not seem predicated on this keep being present - confirmed with "jar tf targets/etc/etc").

### DIFF
--- a/google-http-client/pom.xml
+++ b/google-http-client/pom.xml
@@ -66,9 +66,6 @@
                   <pattern>com.google.common.**</pattern>
                   <result>com.google.api.client.repackaged.com.google.common.@1</result>
                 </rule>
-                <keep>
-                  <pattern>com.google.api.client.**</pattern>
-                </keep>
               </rules>
             </configuration>
           </execution>


ejona86:
The keep reduces the amount of repackaging to only classes that are used. (As mentioned in the [last paragraph of jarjar's readme](https://github.com/shevek/jarjar/blob/master/README.md)).

Oauth should not be directly referencing the repackaged classes. I searched oauth and api, and there were no references to "repackaged".

So I expect that for some reason jarjar is failing to include a class it should. The question is why doesn't it detect the class is necessary. None of http/oauth/api reference Platform. Can you provide the full error message? It should likely make it clear what is using Platform.

jadekler:
Superceded by https://github.com/google/google-http-java-client/pull/434

ejona86:
For posterity, the error was:
```
java.lang.NoClassDefFoundError: com/google/api/client/repackaged/com/google/common/base/Platform
at com.google.api.client.repackaged.com.google.common.base.Strings.isNullOrEmpty(Strings.java:67)
at com.google.api.client.util.Strings.isNullOrEmpty(Strings.java:36)
```

This is with Guava 20 (IIRC). The Paltform reference in Strings is pretty explicit; no real cause for it to be handled improperly. So the thought was it was a jarjar bug. Upgrading the jarjar-maven-plugin fixed the problem. We also discovered that jarjar-maven-plugin forked jarjar :''(

PR #446, opened 2018-08-15
Fix referential comparison of the two strings

chingor13:
Fixes a error-prone-error when importing to internal repo.


chingor13:
We can ignore the Kokoro tests as they aren't configured in this branch. I disabled the webhook until we get the release out.

PR #447, opened 2018-08-16
Fix throwIfFalseEOF if skip method was called before read

medb:
throwIfFalseEOF method don't take into account bytes read/skipped in skip method and throws IOException when legit EOF was reached.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpResponse.java
@@ -186,6 +186,17 @@ final class NetHttpResponse extends LowLevelHttpResponse {
       return n;
     }
 
+    @Override
+    public long skip(long n) throws IOException {
+      int n = in.skip(n);
+      if (n == -1) {
+        throwIfFalseEOF();
+      } else {
+        bytesRead += n;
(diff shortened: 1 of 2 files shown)


chingor13:
Can we add a test for this?

It looks like we can add something like this:

```
  public void testSkippingBytes() throws IOException {
    MockHttpURLConnection connection = new MockHttpURLConnection(null)
        .setResponseCode(200)
        .setInputStream(new ByteArrayInputStream(StringUtils.getBytesUtf8("0123456789")))
        .addHeader("Content-Length", "10");
    NetHttpResponse response = new NetHttpResponse(connection);
    InputStream is = response.getContent();
    // read 1 byte, then skip 9 (to EOF)
    assertEquals('0', is.read());
    assertEquals(9, is.skip(9));
    // expect EOF, not an exception
    assertEquals(-1, is.read());
  }
```

medb:
Thanks for the test, added it to PR

chingor13:
@medb Is it possible to rebase these changes off of the "dev" branch? Strangely, the "dev" branch is the primary branch, not "master" and the CI config in that branch cannot be run.

medb:
Rebased on `dev` branch

chingor13:
Thanks!

PR #454, opened 2018-08-24
Fix compilation for Java 10

chingor13:
* Update datanucleus plugin versions to build on Java 9/10.
* Update clirr-maven-plugin to 2.8 to work on Java 10.
* Removing Java 9 from the test matrix as we cannot resolve the issue and guava does not seem to work on Java 9 anyways.


chingor13:
The Java 9 test was removed internally and should not run on future PRs.

chingor13:
@jadekler The only versions that changed were the maven plugins which aren't included in the assembly - they only are used during the development and build process.

PR #455, opened 2018-08-27
Fix plus signs in URL paths.

chingor13:
When building a URL, store the rawPath. When dealing with pathParts, you
will deal with decoded values, but toString() and getRawPath() deal with
encoded strings.


chingor13:
This did not pass internal testing.

PR #474, opened 2018-09-23
Fix test to run on environments with German locale (#473)

### LINKED ISSUE #473: Test fails on enviroments with German Local
The test ```testContentLoggingLimitWithLoggingEnabledAndDisabled``` fails due to using a fixed decimal separator character (','), whereas the ```LoggingByteArrayOutputStream``` uses ```NumberFormat``` to format the output according to given locale. 

These are the two methods, that make the test fail. ('18,000' needs be adapted)

```
subtestContentLoggingLimit(big, Integer.MAX_VALUE, true, "Total: 18,000 bytes", big);
subtestContentLoggingLimit(big, 4, true, "Total: 18,000 bytes (logging first 4 bytes)", "xxxx");
```

geri-m:
Fixes #473


geri-m:
I signed it!

PR #480, opened 2018-09-24
remove datanucleus

### LINKED ISSUE #478: DataNucleus
See if we can remove datanucleus

elharo:
Fixes #478 @chingor13 

- [X] Tests pass


elharo:
Ping @chingor13

chingor13:
I was holding off because I don't know the implications of removing datanucleus. It rewrites to bytecode to provide the JDO interface, so I'm not sure if we can remove it (or perhaps we no longer want to provide the jdo artifact?).

elharo:
We really, really don't want to be providing or depending on datanucleus in 2018. The version we have here is old, broken, and busted. I don't think there's anywhere we are, but if I missed something please point me at it so I can figure out how to remove or replace it too.

elharo:
Also worth noting that all tests passed while removing these dependencies. Thus if we are depending on this in some way, the test suite is very incomplete.

chingor13:
Correct, we are not using this artifact or JDO features ourselves. `google-http-client-jdo` is an extension utility to provide a JDO interface. DataNucleus is currently enhancing classes in https://github.com/googleapis/google-http-java-client/blob/b4770b5e49554041b0825f5c4d4be50dc732a7cb/google-http-client-jdo/src/main/java/com/google/api/client/extensions/jdo/JdoDataStoreFactory.java#L285 via annotations. We are definitely missing tests to ensure this functionality is present.

I recently updated the data nucleus versions, but we are hamstrung by our support of Java 6. We should be able to update the version again when we drop Java 6 support early 2019.

Is there an alternative to data nucleus that we could consider?

elharo:
What's the context in which this is expected to be used. App Engine recommends the Datastore API or Objectify:

https://cloud.google.com/appengine/docs/standard/java/datastore/jdo/overview-dn2

elharo:
I can find only a couple of usages of this artifact and that's all in old, abandoned, and deprecates code.

PR #485, opened 2018-09-27
WIP: Add write timeout for post/put requests

### LINKED ISSUE #484: POST/PUT data can hang indefinitely and ignore timeouts
It's possible that the local system's output stream blocks when writing data. We should add the ability to set a timeout on the socket write portion. This portion of the request ignores the connect and read timeout settings.

#### Environment details

- OS: Linux
- Java version: 7, 8
- google-http-java-client version: 1.23.0

#### Steps to reproduce

Example test: https://github.com/chingor13/http-write-timeout

#### Any additional information below

A few customers are seeing this issue.

chingor13:
Fixes #484

- [ ] Tests pass

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
@@ -18,9 +18,18 @@ import com.google.api.client.http.LowLevelHttpRequest;
 import com.google.api.client.http.LowLevelHttpResponse;
 import com.google.api.client.util.Preconditions;
 
+import com.google.api.client.util.StreamingContent;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.HttpURLConnection;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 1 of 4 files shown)


chingor13:
We should merge this and build a snapshot for customers to try before we release it.

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java
```diff
@@ -111,4 +127,39 @@ public LowLevelHttpResponse execute() throws IOException {
...
+      final StreamingContent content = getStreamingContent();
+      final Callable<Boolean> writeContent = new Callable<Boolean>() {
+        @Override
+        public Boolean call() throws IOException {
+          content.writeTo(outputStream);
+          return Boolean.TRUE;
+        }
+      };
+      final ExecutorService executor = Executors.newSingleThreadExecutor();
+      final Future future = executor.submit(new FutureTask<Boolean>(writeContent), null);
```
huygaa11: Does creating additional thread affect performance in a negative way?

chingor13: We'll have to investigate deeper. With the current implementation, we are only using the Future if we set a write timeout.

chingor13:
Note that this failed internal testing, so trying to resolve those failures.

PR #506, opened 2018-10-30
Remove deprecated BackOffPolicy interface

chingor13:
It was scheduled to be removed in 1.18 (March 2014)


clirr-ignored-differences.xml
```diff
@@ -7,4 +7,22 @@
     <differenceType>8001</differenceType>
     <className>com/google/api/client/repackaged/**</className>
   </difference>
+  <difference>
```
elharo: Question: why is this being ignored? If clirr is just generating a report, this change should be included. If it's failing the build on changes it should allow for removal of `@Beta`, `@Deprecated` methods?

chingor13: We are using the `check` goal which fails during tests if it detects API breaking changes. Unfortunately, the plugin is not smart enough to understand that we are allowing removal of `@Deprecated` features so we must add it to the ignore list to explicitly acknowledge that we are "breaking" the API.

After the next version, we can bump the version we're comparing against and clear the list (which hasn't been done since 1.19).

chingor13: Note that the first commit in the PR failed CI because the inner `ExponentialBackOffPolicy$Builder` class was not included in the list of classes removed.

PR #523, opened 2018-11-07
Fix maven-javadoc-plugin in pom.xml

### LINKED ISSUE #522: maven-javadoc-plugin with incorrect param
After switching from Version 2.9.1 to 3.0.1 the element ```<additionalparam>-Xdoclint:none</additionalparam>``` needs to be replaced with ```<additionalOptions>-Xdoclint:none</additionalOptions>```. 

See https://issues.apache.org/jira/browse/MJAVADOC-475

geri-m:
Fixes #522 

- [x] Tests pass


igorbernstein2:
Won't this break jdk7 builds? `-Xdoclint` is only available starting from jdk8

chingor13:
Hmm, it didn't seem to break the Java 7 build here. I'll investigate further.

geri-m:
Actually without this change, the Jdk8 build on my local machine was not going thru, which was wired as well.

geri-m:
I justed check the CI Script. Currently 
```
mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
mvn test
``` 

is what CI executes. JavaDoc is skipped, hence CI for any JDK would work. Should this be changed?

igorbernstein2:
I believe the correct thing to do would be to wrap the Xdoclint options in a profile that is activated for jdk >= 8

chingor13:
Filed #524

PR #529, opened 2018-11-21
Fix int type transformed as BigDecimal value when parsing as Map

ebuildy:
Using with Jackson2 parser,  a simple parsing of { "data" : 1} gives a BigDecimal instead an integer. In my case, ``valueClass`` is Object, but the JSON current token is VALUE_NUMBER_INT not FLOAT.

I parse as a Map. No problem if I parse as a class with "int" type.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/json/JsonParser.java
+++ b/google-http-client/src/main/java/com/google/api/client/json/JsonParser.java
@@ -828,6 +828,9 @@ public abstract class JsonParser implements Closeable {
           Preconditions.checkArgument(
               fieldContext == null || fieldContext.getAnnotation(JsonString.class) == null,
               "number type formatted as a JSON number cannot use @JsonString annotation");
+          if (getCurrentToken() == VALUE_NUMBER_INT && (valueClass == null || valueClass.isAssignableFrom(Integer.class))) {
+            return getIntValue();
+          }
           if (valueClass == null || valueClass.isAssignableFrom(BigDecimal.class)) {
             return getDecimalValue();
           }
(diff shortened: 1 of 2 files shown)


ebuildy:
I signed it!

sduskis:
@ebuildy, can you please add a unit test for this change?

ebuildy:
Sure I will this week, thanks you.

sduskis:
@ajaaym, this needs a unit test.  Can you please get someone to resubmit this with a unit test?

ajaaym:
@sduskis yes

pmakani:
@sduskis  unit test added, PTAL.

google-http-client/src/main/java/com/google/api/client/json/JsonParser.java
```diff
@@ -32,31 +24,22 @@
...
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
-import java.util.WeakHashMap;
+import java.util.*;
```
sduskis: Please use IDE settings that do not use `*` but rather use individual imports.

sduskis:
@pmakani, there are a lot of formatting changes here.  Please raise a separate PR for just the formatting changes.

pmakani:
> @pmakani, there are a lot of formatting changes here. Please raise a separate PR for just the formatting changes.

PR #619

chingor13:
Turns out this behavior change is breaking a bunch of internal customers.

PR #533, opened 2018-12-04
Fix GenericUrl to decode path parts correctly

### LINKED ISSUE #398: GenericUrl converts plus sign into space within URI path component
```java
    @Test

    }
```

```xml
        <dependency>
            <version>1.23.0</version>
        </dependency>
```

ajaaym:
Fixes #398  GenericUrl converts plus sign into space within URI path component

- [ ] Tests pass


chingor13:
This one is actually non-trivial because it breaks many internal targets (which is why it's been fixed, then reverted several times).

ajaaym:
@chingor13 ok thanks. In that case should we put back label ```status: will not fix``` in the issue?

PR #541, opened 2018-12-11
Deprecate AndroidHttp compatibility shim

chingor13:
This is a beta API that should no longer be necessary. Gingerbread is no longer supported by Google Play Services so this could only be useful for someone building an extremely legacy app (and they would need to use an earlier version anyways).


elharo:
Can we remove this completely in the next release?

Issue #494, opened 2018-10-16
Re-add PATCH to apache adapter


chingor13:
We reverted #486 in #493 to unblock a release. We need to re-add this method after we verify upgrades internally.

chingor13:
This is currently blocked on internal dependencies.

chingor13:
Fixed the internal blocking dependency.

PR #543, opened 2018-12-11
Split http apache artifact

chingor13:
This should allow us to maintain a legacy apache http artifact for compatibility while we can have external users use the latest apache http client.

This will enable us to handle #494 

The issue now is that users who want to use the apache adapter will need to import a new artifact: `google-http-client-apache`.


loosebazooka:
Should the dependency be moved as well (into the sub-projects)? https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/pom.xml#L166

PR #545, opened 2018-12-12
Re-add OpenCensus integration

### LINKED ISSUE #459: Add OpenCensus tracing
OpenCensus integration was reverted prior to the 1.24.1 release.

chingor13:
Fixes #459

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
@@ -25,6 +25,8 @@ import com.google.api.client.util.StreamingContent;
 import com.google.api.client.util.StringUtils;
 
 import io.opencensus.common.Scope;
+import io.opencensus.contrib.http.util.HttpTraceAttributeConstants;
+import io.opencensus.trace.AttributeValue;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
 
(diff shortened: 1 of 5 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -883,7 +891,12 @@ public HttpResponse execute() throws IOException {
     Preconditions.checkNotNull(requestMethod);
     Preconditions.checkNotNull(url);
 
+    Span span = tracer
+        .spanBuilder(OpenCensusUtils.SPAN_NAME_HTTP_REQUEST_EXECUTE)
```
ajaaym: Shouldn't this use resource name as span name ?

chingor13: This span is meant to identify the underlying http request being made. The caller (e.g. in google-cloud-java or whatever library uses google-http-client) should also have a span that identifies the resource/method being called. This span will be a child of that parent span.

sduskis: The java grpc implementation puts in "Sent.<operationName>" as the span, without any assumption about a parent span.  It would be great if the HTTP implementation could also have a span name that has some context about the call, like the URL

chingor13:
The title was correct, this was previously merged, and then reverted due to time crunch on integrating internally.

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
chingor13: @bogdandrutu Is it preferable to have the url as the span name here or as an attribute?

bogdandrutu:
If possible avoid a high cardinality span name. If the URL does not have a high cardinality and it doesn't use any PII (like resource name) data then should be fine.

bogdandrutu:
See this document https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
chingor13: Ok, as this is a general-use client that is can be and is used outside the scope of our control, we should not include the URL by default as we can't guarantee that the URL does not contain secrets. We should ensure that our higher level clients set the URL or route as the span name.

sduskis: Is it safe to add an attirbute for the URL?

chingor13: Adding as an attribute solves the cardinality issue, but not hiding PII. Users don't have control over this span, so they wouldn't be able opt out.

bogdandrutu: I think adding it as an attribute is good starting point. If somebody complains about that we can reconsider.

PR #548, opened 2018-12-28
Fixed MockBackOff#getMaxTries, returns `maxTries`

sebfz1:
Fixes `MockBackOff#getMaxTries` which should return `maxTries` (instead of `numTries`)  

- [x] Tests pass


chingor13:
Thanks!

PR #550, opened 2018-12-28
Fix iterable maps JSON serialization

chingor13:
From #404 

If a map is iterable, then the JSON generator treats the map as an array when it should still be treated as a JSON object.


chingor13:
@mikekap signed the CLA for that commit in #404

mikekap:
Thanks for getting this one over the line @chingor13 !

PR #557, opened 2019-01-03
Use NIO to set file permissions

### LINKED ISSUE #315: FileDataStoreFactory method setPermissionsToOwnerOnly only supports Unix
com.google.api.client.util.store.FileDataStoreFactory

setPermissionsToOwnerOnly does not use newer Java 7 or 8 APIs for setting file/dir ownership and thus generates spurious warnings like "com.google.api.client.util.store.FileDataStoreFactory setPermissionsToOwnerOnly
WARNING: unable to change permissions for everybody: "

The code only takes into account file systems as seen on Unix-like OSes that use a simple mode for the permissions.   Windows file systems have ACLs and require different API calls to modify ownership and permissions.  The latest Java APIs should be used instead which can handle both file system types.

This warning occurs for anyone using the Drive API SDK samples and causes confusion for new developers using the API.

https://github.com/google/google-http-java-client/blob/dev/google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java

chingor13:
Fixes #315 

We're dropping Java 6 support in this next release, so we can now use NIO to set file permissions in Windows environments


chingor13:
Adding windows test runner in #562

acurvers:
why did this not get merged?

chingor13:
This was not merged as we were waiting until we set up a Windows test machine. This is blocked by #716

PR #558, opened 2019-01-03
Update ApacheHttpTransport implementation

### LINKED ISSUE #364: AndroidEnterpriseClient doesn't work with HttpClient instance obtained from HttpClients.customs method
CloseableHttpClient closeableHttpClient = HttpClients.custom()...build();
ApacheHttpTransport apacheHttpTransport = new ApacheHttpTransport(closeableHttpClient);
AndroidEnterprise enterpriseClient = new AndroidEnterprise.Builder(apacheHttpTransport, ...);
AndroidEnterprise.Products.List listProducts = enterpriseClient.products().list(enterpriseId);

The above piece of code throws the following exception.
java.lang.UnsupportedOperationException
	at org.apache.http.impl.client.InternalHttpClient.getParams(InternalHttpClient.java:210)
	at com.google.api.client.http.apache.ApacheHttpTransport.<init>(ApacheHttpTransport.java:129)

### LINKED ISSUE #482: Apache HttpClient is not working with ApacheHttpTransport implementation
Creating ApacheHttpTransport using HttpClient constructor argument is throwing java.lang.UnsupportedOperationException

```
public static HttpTransport getHttpTransport() {
            final PoolingHttpClientConnectionManager httpClientConnectionManager = new PoolingHttpClientConnectionManager(CONNECTION_TTL_MILLIS,
                    TimeUnit.MILLISECONDS);
            httpClientConnectionManager.setDefaultMaxPerRoute(TOTAL_MAX_CONNECTIONS_PER_ROUTE);
            httpClientConnectionManager.setMaxTotal(TOTAL_MAX_CONNECTIONS);
            httpClientConnectionManager.setValidateAfterInactivity(INACTIVE_CONNECTION_VALIDATION_MILLIS);
            final RequestConfig requestConfig = RequestConfig.custom()
                    .setConnectTimeout(CONNECT_TIMEOUT_MILLLIS)
                    .setSocketTimeout(SOCKET_TIMEOUT_MILLIS)
                    .build();
            final CloseableHttpClient httpClient = HttpClientBuilder.create()
                    .setDefaultRequestConfig(requestConfig)
                    .setConnectionManager(httpClientConnectionManager)
                    .build();
            return new ApacheHttpTransport(httpClient);
        }
        public static void main(String args[]) {
            ApacheHttpClientUtil.getHttpTransport();
        }
```

#### Environment details

- OS: Mac 10.13.5 (17F77)
- Java version: 1.8.0_161
- google-http-java-client version: 1.25.0

#### Steps to reproduce

Create ApacheHttpTransport object using constructor which takes HttpClient object. 
HttpClient is created using HttpClientBuilder. 

#### Stacktrace

```
Caused by: java.lang.UnsupportedOperationException: null
	at org.apache.http.impl.client.InternalHttpClient.getParams(InternalHttpClient.java:211)
	at com.google.api.client.http.apache.ApacheHttpTransport.<init>(ApacheHttpTransport.java:126)
	at com.mobileiron.corona.efa.config.FireBaseConfiguration$ApacheHttpClientUtil.getHttpTransport(FireBaseConfiguration.java:133)
	at com.mobileiron.corona.efa.config.FireBaseConfiguration.<init>(FireBaseConfiguration.java:47)
	at com.mobileiron.corona.efa.config.FireBaseConfiguration$$EnhancerBySpringCGLIB$$a638c5f0.<init>(<generated>)
	at sun.reflect.GeneratedConstructorAccessor437.newInstance(Unknown Source)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.springframework.beans.BeanUtils.instantiateClass(BeanUtils.java:170)
	... 88 common frames omitted
```

It looks like HttpClientBuilder is creating instance of InternalHttpClient which throws UnsupportedOperationException when calling getParams method. Since many people use apache client's HttpClientBuilder it will good to fix the issue. 
```
Most of the APIs(APIs from Apache Http Client) used by ApacheHttpTransport are deprecated.
```

### LINKED ISSUE #503: ApacheHttpTransport uses deprecated APIs and cannot be used with recommended HttpClient APIs
`ApacheHttpTransport` class uses many deprecated APIs from Apache HTTP Client. There are around 75 deprecation warnings in that class alone.

Moreover, it is now impossible to use the `ApacheHttpTransport` with a recommended implementation of Apache HttpClient. For instance, the following doesn't work:

```
import org.apache.http.impl.client.HttpClients;

public void testDefaultClient() {
  new ApacheHttpTransport(HttpClients.createDefault());
}
```

The above fails with the following error:

```
java.lang.UnsupportedOperationException
	at org.apache.http.impl.client.InternalHttpClient.getParams(InternalHttpClient.java:211)
	at com.google.api.client.http.apache.ApacheHttpTransport.<init>(ApacheHttpTransport.java:127)
	at com.google.api.client.http.apache.ApacheHttpTransportTest.testDefaultClient(ApacheHttpTransportTest.java:98)
```

This is because new `HttpClient` implementations do not expose their `HttpParams` and `ConnectionManager` (both these getters have been deprecated, and most default implementations now throw errors).

Realistically one can only use `ApacheHttpTransport` with `DefaultHttpClient`, which itself is a deprecated API now.

chingor13:
Fixes #364 
Fixes #482 
Fixes #503 

This rewrites the `ApacheHttpTransport` to use non-deprecated functionality. In the upgrade, we lose ability to modify clients after the fact - they became read-only. Thus, we cannot enforce things like disabling retries on provided `HttpClient` implementations.

We now have 2 artifacts:
1. The "legacy" one in `google-http-client-apache-legacy/` which is the old one exactly which only works with the default implementation. This one will be available as version 1.28.0.
2. The "new" one in `google-http-client-apache/` which "removes" functionality like the builder - users should opt for the builder provided by Apache HttpClient. It has breaking changes, so this one will be available as version 2.0.0.


google-http-client-apache-legacy/pom.xml
```diff
@@ -7,9 +7,9 @@
     <version>1.27.1-SNAPSHOT</version><!-- {x-version-update:google-http-client-parent:current} -->
     <relativePath>../pom.xml</relativePath>
   </parent>
-  <artifactId>google-http-client-apache-legacy</artifactId>
+  <artifactId>google-http-client-apache</artifactId>
```
kolea2: why remove the "-legacy"?

google-http-client-apache/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
```diff
@@ -234,7 +186,13 @@ protected ApacheHttpRequest buildRequest(String method, String url) {
    */
   @Override
   public void shutdown() {
-    httpClient.getConnectionManager().shutdown();
+    if (httpClient instanceof CloseableHttpClient) {
+      try {
+        ((CloseableHttpClient) httpClient).close();
+      } catch (IOException e) {
+        // ignore
```
kolea2: why ignore?

google-http-client-apache-legacy/pom.xml
chingor13: We haven't yet published either artifact yet. Originally, I had planned for the new version to be backwards compatible, so we would keep 2 separate artifacts at the same version. However, after it became clear that we needed to have breaking changes to support the "new" Apache HttpClient implementations, it made more sense to have the same artifact with 2 major versions.

This is totally up for discussion though.

google-http-client-apache/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
chingor13: Ah, close can throw an `IOException` - I'll let it bubble up.

chingor13: Done.

PR #562, opened 2019-01-04
Add Java 8 on windows test config

chingor13:

PR #563, opened 2019-01-04
Remove commons-codec dependency

### LINKED ISSUE #457: Remove commons-codec dependency
We use commons-codec for:
- [ ] base64 encoding/decoding
- [ ] String byte utils

chingor13:
Fixes #457 

Use guava Base64 handling until we drop Java 7 support and can use the built-in classes.
Use Java 7 charset handling for StringUtils.

By removing commons-codec, we should no longer need to jarjar a shaded version of commons-codec (and guava). We can also remove the antrun plugin which scrubbed the shaded commons-codec. This will also unblock the windows tests in #562 (file permissions errors during antrun plugin) Edit: it unblocked it, but there are more failures.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/StringUtils.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/StringUtils.java
@@ -15,18 +15,11 @@
 package com.google.api.client.util;
 
 import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
 
 
 /**
  * Utilities for strings.
- *
- * <p>
- * Some of these methods are a proxy for version 1.6 (or newer) of the Apache Commons Codec
- * {@link StringUtils} implementation. This is needed in order to support platforms like Android
- * which already include an older version of the Apache Commons Codec (Android includes version
- * 1.3). To avoid a dependency library conflict, this library includes a reduced private copy of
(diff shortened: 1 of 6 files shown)


google-http-client/src/main/java/com/google/api/client/util/StringUtils.java
```diff
@@ -48,13 +41,15 @@
...
    *         according the the Java specification.
-   * @see <a href="http://download.oracle.com/javase/1.5.0/docs/api/java/nio/charset/Charset.html"
+   * @see <a href="http://download.oracle.com/javase/7/docs/api/java/nio/charset/Charset.html"
    *      >Standard charsets</a>
-   * @see org.apache.commons.codec.binary.StringUtils#getBytesUtf8(String)
    * @since 1.8
    */
   public static byte[] getBytesUtf8(String string) {
-    return org.apache.commons.codec.binary.StringUtils.getBytesUtf8(string);
+    if (string == null) {
```
kolea2: Would you mind adding a small test for this and newStringUtf8 please?

kolea2: (unless this method is already tested somewhere else?)

chingor13: Added, PTAL.

PR #567, opened 2019-01-07
Move OpenCensusUtils to com.google.api.client.http

chingor13:
Avoid circular dependency between packages.
Also makes the class default visibility as it's an internal class. This should be fine because it hasn't been released yet.


google-http-client/src/main/java/com/google/api/client/http/OpenCensusUtils.java
```diff
@@ -42,7 +40,7 @@
  * @author Hailong Wen
  * @since 1.28
  */
-public class OpenCensusUtils {
+class OpenCensusUtils {
```
kolea2: should this be private?

PR #585, opened 2019-02-10
Revert "remove datanucleus (#480)"

elharo:
@chingor13 Noticed the do not merge tag too late, so if you do want to roll the removal of datanucleus back (I still suspect we don't need it) here's a PR.


sduskis:
We can restore it at some other point.

PR #588, opened 2019-02-12
372: Changed SingleThreadExecutor to FixedThreadPool(1).

### LINKED ISSUE #372: HttpRequest#executeAsync() may throw a RejectedExecutionException
As described in the question and answer in [this Stack Overflow post](https://stackoverflow.com/questions/46457678/garbage-collection-seems-to-shutdown-local-executor-and-cause-rejectedexecutione) and [this JDK issue](https://bugs.openjdk.java.net/browse/JDK-8145304), the `ExecutorService` returned by `Executors.newSingleThreadExecutor` can be finalized and shut down while still in use.

This may cause a `RejectedExecutionException` to be thrown while [`HttpRequest#executeAsync(Executor)`](https://github.com/google/google-http-java-client/blob/dev/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java#L1094) is submitting its task.

I suggest changing `HttpRequest#executeAsync()` to use an `ExecutorService` returned by `Executors#newFixedThreadPool(1)`, which isn't wrapped in a `FinalizableDelegatingExecutorService`.

andrey-qlogic:
Fixes #372

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
@@ -1200,14 +1200,14 @@ public final class HttpRequest {
   /**
    * {@link Beta} <br/>
    * Executes this request asynchronously using {@link #executeAsync(Executor)} in a single separate
-   * thread using {@link Executors#newSingleThreadExecutor()}.
+   * thread using {@link Executors#newFixedThreadPool(1)}.
    *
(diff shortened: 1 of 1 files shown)


sduskis:
Does it make sense to use a daemon threads so that the requests do not block jvm shutdown?

andrey-qlogic:
@sduskis , good point, thanks

PR #589, opened 2019-02-12
Implement equals and Hashcode on GenericData

andrey-qlogic:
Fixes [#936](https://github.com/googleapis/google-api-java-client/issues/936)

- [ X] Tests pass

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/GenericData.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/GenericData.java
@@ -204,26 +204,19 @@ public class GenericData extends AbstractMap<String, Object> implements Cloneabl
     }
     if (o instanceof GenericData) {
       GenericData that = (GenericData) o;
-      return Objects.equals(this.classInfo, that.classInfo)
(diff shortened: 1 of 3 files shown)


google-http-client/src/main/java/com/google/api/client/util/GenericData.java
```diff
@@ -196,6 +197,28 @@ public final void setUnknownKeys(Map<String, Object> unknownFields) {
...
+  @Override
+  public boolean equals(Object o) {
+    if (o == this) {
+      return true;
+    }
+    if (o instanceof GenericData) {
+      GenericData that = (GenericData) o;
+      return super.equals(that) && Objects.equals(this.classInfo, that.classInfo);
+    }
+    return false;
```
sduskis: I think this is better:

```
    if ((!o instanceof GenericData)) {
      return false;
    }
    GenericData that = (GenericData) o;
    return super.equals(that) && Objects.equals(this.classInfo, that.classInfo);
```

andrey-qlogic: @sduskis thanks for the comment. Done.

PR #591, opened 2019-02-13
depend directly on Apache httpclient

elharo:
depend directly on Apache httpclient instead of pulling it in transitively from weird android dependency.

- [X] Tests pass

Not sure why tests didn't run, but there's a local failure I'm investigating.

OK.  tests passed locally. First failure was probably a DNS glitch caused by putting laptop to sleep in the middle. 

@sduskis I'm still concerned that the kokoro jobs didn't fire here.


elharo:
@sduskis Do you know how I can manually trigger the kokoro builds?

elharo:
OK, looks there's a kokoro:force-run label

PR #599, opened 2019-02-18
Added method getContentAsInputStream.

### LINKED ISSUE #491: HttpResponseException.getContent() should return an InputStream
#### Environment details

- OS: Win 10
- Java version: 8 u144
- google-http-java-client version:  1.25.0

#### Steps to reproduce

1. Create a request for the URL: `https://maps.googleapis.com/maps/api/staticmap?markers=-30.93963400,-61.56805900&size=672x297`
2. It will throw a `HttpResponseException` since the page returns a `403 Forbidden`

The problem here is that `HttpResponseException.getContent()` returns a `String` (implemented for #143). However, the response for that URL is an image (`content-type: image/png`).

I think that `HttpResponseException.getContent()` should be equal to `HttpResponse.getContent()`. It is currently being parsed at [HttpResponseException.java#L175](https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java#L175).

andrey-qlogic:
Fixes #491

- [x] Tests pass


ajaaym:
@sduskis That would be breaking change and this normally doesn't happen for google services. Thats why @chingor13 suggested to add new method which returns input stream.

andrey-qlogic:
@sduskis, PTAL

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -127,6 +129,16 @@ public final String getContent() {
     return content;
   }
 
+  /**
+   * Returns the HTTP response or {@code null} for none as an input stream.
+   *
+   * @since 1.28
+   */
+  public final InputStream getContentAsInputStream() {
+    if (content == null) return null;
```
elharo: nit: google style guide always uses parentheses for if and other block statements

andrey-qlogic: Resolved

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -127,6 +129,18 @@ public final String getContent() {
...
+  /**
+   * Returns the HTTP response or {@code null} for none as an input stream.
+   *
+   * @since 1.28
+   */
+  public final InputStream getContentAsInputStream() {
+    if (content == null) {
+      return null;
+    }
+    return new ByteArrayInputStream(content.getBytes());
```
sduskis: This doesn't seem right.  Please confirm that this works with an image, or some other binary values.

andrey-qlogic: The change uses ByteArrayInputStream  because the Builder that builds the content uses the HttpResponse method parseAsString 
https://github.com/googleapis/google-http-java-client/blob/380a60b499b845a4ba423e1105a469dabfdf4ad2/google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java#L175

https://github.com/googleapis/google-http-java-client/blob/a7ee06950b8309787127340e90be0b528c197378/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java#L514
which uses ByteArrayOutputStream 

The change just reverts the string back to the input stream. Please correct me if I'm wrong.

ajaaym: @andrey-qlogic This is not exactly correct it will not convert back to bytes correctly if charset is not same and error prone too. Can you please make change to store original content in to byte and then when somebody calls getContent convert to string and when asking for stream pass it as stream?

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -267,31 +289,47 @@ public HttpHeaders getHeaders() {
...
     public Builder setHeaders(HttpHeaders headers) {
       this.headers = Preconditions.checkNotNull(headers);
       return this;
     }
 
     /** Returns the HTTP response content or {@code null} for none. */
-    public final String getContent() {
-      return content;
+    public final String getContent() throws IOException {
+      if (content == null || charset == null) {
```
ajaaym: why charset == null? if charset is not available it should use default charset.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: Done

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: Removed this check

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -123,8 +131,22 @@ public HttpHeaders getHeaders() {
    *
    * @since 1.14
    */
-  public final String getContent() {
-    return content;
+  public final String getContent() throws IOException {
+    if (content == null) {
+      return null;
+    }
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
```
sduskis: Does this work `return new String(content, charset.name())`?

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -172,9 +197,9 @@ public Builder(HttpResponse response) {
       this(response.getStatusCode(), response.getStatusMessage(), response.getHeaders());
       // content
       try {
-        content = response.parseAsString();
-        if (content.length() == 0) {
-          content = null;
+        InputStream inputStream = response.getContent();
+        if (inputStream != null) {
+          content = ByteStreams.toByteArray(response.getContent());
```
sduskis: nit: this should just be `ByteStreams.toByteArray(inputStream)`

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -123,8 +131,22 @@ public HttpHeaders getHeaders() {
...
+    ByteStreams.copy(new ByteArrayInputStream(content), out);
+    return out.toString(charset.name());
+  }
+
+  /**
+   * Returns the HTTP response or {@code null} for none as an input stream.
+   *
+   * @since 1.28
+   */
+  public final InputStream getContentAsInputStream() {
```
sduskis: I think that we ought to be more direct here and just do this:

```
public final byte[] getRawContent() {
   return content;
}
```

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -185,7 +210,19 @@ public Builder(HttpResponse response) {
       // message
       StringBuilder builder = computeMessageBuffer(response);
       if (content != null) {
-        builder.append(StringUtils.LINE_SEPARATOR).append(content);
+        try {
+          // charset
+          charset = response.getContentCharset();
```
sduskis: `getContentCharset()` doesn't throw an exception.  Please move it up into the ` if (inputStream != null)` section.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -123,8 +131,22 @@ public HttpHeaders getHeaders() {
    *
    * @since 1.14
    */
-  public final String getContent() {
-    return content;
+  public final String getContent() throws IOException {
```
sduskis: IOException cannot be throw here.  That's backwards incompatible.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: It can throw as it uses from Charset.java 
public static Charset forName(String charsetName) {
        Charset cs = lookup(charsetName);
        if (cs != null)
            return cs;
        throw new UnsupportedCharsetException(charsetName);
    }

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: Removed

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: This is a good point

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: Yes, is is also work

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: thanks

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -124,6 +130,24 @@ public HttpHeaders getHeaders() {
...
    */
   public final String getContent() {
+    if (content == null) {
+      return null;
+    }
+    try {
+      return new String(content, charset.name());
+    } catch (UnsupportedEncodingException exception) {
+      // it would be bad to throw an exception while throwing an exception
+      exception.printStackTrace();
```
sduskis: Is there some RuntimeException we can throw?

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -172,9 +199,9 @@ public Builder(HttpResponse response) {
       this(response.getStatusCode(), response.getStatusMessage(), response.getHeaders());
       // content
       try {
-        content = response.parseAsString();
-        if (content.length() == 0) {
-          content = null;
+        InputStream inputStream = response.getContent();
+        if (inputStream != null) {
```
sduskis: Please add a comment about the fact that we are reading the input stream so that the connection closes.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -185,7 +212,19 @@ public Builder(HttpResponse response) {
...
-        builder.append(StringUtils.LINE_SEPARATOR).append(content);
+        try {
+          // charset
+          charset = response.getContentCharset();
+          String strContent = getContent();
+          if (strContent != null) {
+            builder.append(StringUtils.LINE_SEPARATOR).append(strContent);
+          }
+        } catch (IOException exception) {
+          // it would be bad to throw an exception while throwing an exception
```
sduskis: We should do a log statement rather than just a print stack trace, if there are other examples of logging.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -253,31 +292,45 @@ public HttpHeaders getHeaders() {
...
+      return this;
+    }
+
+    /**
+     * Sets the HTTP response charset or {@code null} for none.
+     *
+     * <p>Overriding is only supported for the purpose of calling the super implementation and
+     * changing the return type, but nothing else.
+     */
+    public Builder setCharset(Charset charset) {
```
sduskis: Do we need this to be public?

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
ajaaym: @sduskis no, we are already in exception case

sduskis: Can we log instead?

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: @sduskis, there is no examples and this library is not logging.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: @sduskis, there is no examples and this library is not logging.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: @sduskis, the public access to this method used in HttpResponseExceptionTest class for unit tests.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: Added comment.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
elharo: For a library we really shouldn't be calling printStackTrace. Maybe we can verify the charset sooner?

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -185,7 +213,19 @@ public Builder(HttpResponse response) {
...
+          charset = response.getContentCharset();
+          String strContent = getContent();
+          if (strContent != null) {
+            builder.append(StringUtils.LINE_SEPARATOR).append(strContent);
+          }
+        } catch (IOException exception) {
+          // it would be bad to throw an exception while throwing an exception
+          exception.printStackTrace();
+        } catch (IllegalArgumentException exception) {
+          exception.printStackTrace();
```
elharo: ditto with above: libraries shouldn't print

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
sduskis: If this is used only in tests, can we mark this package private (i.e. no public/protected/private modifier), and add the `@VisibleForTesting` annotation?

elharo: Then non-public and @VisibleForTesting

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -253,31 +293,45 @@ public HttpHeaders getHeaders() {
...
-     * Overriding is only supported for the purpose of calling the super implementation and changing
-     * the return type, but nothing else.
-     * </p>
+     * <p>Overriding is only supported for the purpose of calling the super implementation and
+     * changing the return type, but nothing else.
      */
     public Builder setContent(String content) {
-      this.content = content;
+      if (content != null) {
+        this.content = content.getBytes();
```
elharo: Should this set the content to null or perhaps empty if null?

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -253,31 +293,45 @@ public HttpHeaders getHeaders() {
...
+
+    /**
+     * Sets the HTTP response charset or {@code null} for none.
+     *
+     * <p>Overriding is only supported for the purpose of calling the super implementation and
+     * changing the return type, but nothing else.
+     */
+    public Builder setCharset(Charset charset) {
+      if (charset != null) {
+        this.charset = charset;
```
elharo: Should this set the content to null or perhaps empty if null?

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: @elharo 
the class prints exception stacktrace, trying to be consistant
https://github.com/googleapis/google-http-java-client/blob/8217035a79bfdc0f4a7193293a8583ee831ee25a/google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java#L171

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: @elharo as a class field byte[] content; is null by initialization, I would assume that to make it empty when the string is null make no sense because the string could be empty.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: @elharo, 
as a class field Charset charset; is null by initialization, why it should be empty if the parameter is null?

andrey-qlogic:
@sduskis , PTAL

sduskis:
I think that we need to stop this investigation.  There doesn't seem to be an easy solution to this issue.

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
sduskis: Please just return null.  We can't print a stack trace.

andrey-qlogic: Done. Removed printing stacktrace. Returns null at the end of the method

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: Done

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
andrey-qlogic: Done

andrey-qlogic:
@sduskis  IllegalCharsetNameException does not have message, may be we need write our own message to a builder ?

sduskis:
Adding "do not merge", pending a review from @chingor13

google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -245,8 +296,11 @@ public Builder setHeaders(HttpHeaders headers) {
     }
 
     /** Returns the HTTP response content or {@code null} for none. */
-    public final String getContent() {
-      return content;
+    public final String getContent() throws IOException {
```
chingor13: This is possibly a breaking change of the API as this builder class is public for some reason. It's fine for binary compatibility but can break source compatibility.

sduskis:
@chingor13, I'm closing this.  We can chat offline.

PR #601, opened 2019-02-19
Added Warning header and unit tests.

### LINKED ISSUE #355: Add support for Warning headers?
It would be useful to support:

https://tools.ietf.org/html/rfc7234#section-5.5

andrey-qlogic:
Fixes #355

- [x] Tests pass

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
@@ -816,14 +816,14 @@ public class HttpHeaders extends GenericData {
   }
 
   /**
-   * Sets the {@code "Warning"} header or {@code null} for none.
+   * Adds the {@code "Warning"} header or {@code null} for none.
    *
    * <p>Overriding is only supported for the purpose of calling the super implementation and
    * changing the return type, but nothing else.
    *
    * @since 1.28
    */
-  public HttpHeaders setWarning(String warning) {
+  public HttpHeaders addWarning(String warning) {
     if (warning == null) {
       return this;
     }
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
@@ -824,17 +824,15 @@ public class HttpHeaders extends GenericData {
    * @since 1.28
    */
   public HttpHeaders setWarning(String warning) {
-    this.warning = getAsList(warning);
+    if (warning == null) {
+      return this;
+    }
+    if (this.warning == null) {
+      this.warning = getAsList(warning);
+    } else {
+      this.warning.add(warning);
+    }
     return this;
-  }
-
-  /**
-   * Returns the first {@code "Warning"} header or {@code null} for none.
-   *
-   * @since 1.28
-   */
-  public final String getWarning() {
-    return getFirstHeaderValue(warning);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
```diff
@@ -811,6 +815,37 @@ public HttpHeaders setAuthenticate(String authenticate) {
     return this;
   }
 
+  /**
+   * Sets the {@code "Warning"} header or {@code null} for none.
+   *
+   * <p>Overriding is only supported for the purpose of calling the super implementation and
```
elharo: Is this enforced in code? Could it be, perhaps with a helper method that's protected and making this method final? But that wouldn't allow you to change the return type I suppose.

google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
```diff
@@ -811,6 +815,37 @@ public HttpHeaders setAuthenticate(String authenticate) {
...
+  public final String getWarning() {
+    return getFirstHeaderValue(warning);
+  }
+
+  /**
+   * Returns all {@code "Warning"} headers or {@code null} for none.
+   *
+   * @since 1.28
+   */
+  public final List<String> getWarningAsList() {
```
elharo: Does this expose potentially mutable internal state?

google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
```diff
@@ -811,6 +815,37 @@ public HttpHeaders setAuthenticate(String authenticate) {
...
+  /**
+   * Sets the {@code "Warning"} header or {@code null} for none.
+   *
+   * <p>Overriding is only supported for the purpose of calling the super implementation and
+   * changing the return type, but nothing else.
+   *
+   * @since 1.28
+   */
+  public HttpHeaders setWarning(String warning) {
+    this.warning = getAsList(warning);
```
elharo: Can there be more than one of these? The code seems undecided.

google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
andrey-qlogic: Done.

google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
```diff
@@ -811,6 +815,35 @@ public HttpHeaders setAuthenticate(String authenticate) {
...
 
+  /**
+   * Sets the {@code "Warning"} header or {@code null} for none.
+   *
+   * <p>Overriding is only supported for the purpose of calling the super implementation and
+   * changing the return type, but nothing else.
+   *
+   * @since 1.28
+   */
+  public HttpHeaders setWarning(String warning) {
```
elharo: Looks like `addWarning` would be more accurate than `setWarning`.

andrey-qlogic: @elharo , thanks for the comment, renamed method.

andrey-qlogic:
@sduskis , PTAL

google-http-client/src/main/java/com/google/api/client/http/HttpHeaders.java
```diff
@@ -158,6 +158,10 @@ public HttpHeaders() {
   @Key("User-Agent")
   private List<String> userAgent;
 
+  /** {@code "Warning"} header. */
+  @Key("Warning")
+  private List<String> warning;
```
sduskis: please rename this variable to `warnings`.

andrey-qlogic: Renamed

PR #614, opened 2019-03-14
move apache classes back into google-http-client

### LINKED ISSUE #606: Fix backwards compatibility in google-api-client
#543 broke backwards compatibility.  (see [here](https://github.com/googleapis/google-cloud-java/issues/4579) for the ripple effect).

#543 broke apart `google-api-client` into two packages, but removed some public classes.  That type of change requires a major semver change.

In the short term, we need to add a dependency in `google-http-client` to `google-http-client-apache` to preserve backwards compatibility until we have a major new release of `google-http-client`.

elharo:
Fixes #606
- [X] Tests pass

I think this reverts the bare minimum files to keep the API compatible with 1.27.0. I left the tests and other files moved or created in #543 behind in google-http-client-apache.


sduskis:
Its weird that we have eclipse files checked in, IMHO.

elharo:
[WARNING] 6003: com.google.api.client.http.HttpRequest: Value of compile-time constant USER_AGENT_SUFFIX has been changed
[WARNING] 6003: com.google.api.client.http.HttpRequest: Value of compile-time constant VERSION has been changed
[ERROR] 7006: com.google.api.client.http.apache.ApacheHttpTransport: Return type of method 'public org.apache.http.impl.client.DefaultHttpClient newDefaultHttpClient()' has been changed to org.apache.http.client.HttpClient

elharo:
We can probably delete and .gitignore those Eclipse files in another PR.

elharo:
The return type change here is weird. Funny it didn't show up before. But it is in a package private method:

https://developers.google.com/api-client-library/java/google-http-java-client/reference/1.19.0/?com/google/api/client/util/class-use/LoggingByteArrayOutputStream.html

elharo:
Looks like the return type of `newDefaultHttpClient` changed in #558 after the artifacts were split so that change needs to be rolled back too. 

https://github.com/googleapis/google-http-java-client/commit/3d7f17c1acd5d8d6e50488ee7c6d047ba614ec11

sduskis:
Is it possible to do the two different reverts in two steps?  I'm having cognitive overload with the combination of the two changes.

elharo:
OK, I've reverted back to the state where it just moved the classes back into the main artifact. This will need a second PR to undo the return type change from #558.

sduskis:
retesting java 7

PR #620, opened 2019-03-21
remove legacy apache module

### LINKED ISSUE #618: Update Apache Http Client in 1.29.0?
The 1.28.0-update for the legacy Google-Apache-Http-Client downgraded it to 4.2.6.
Was this downgrade really neccessary?
According to snyk.io, the downgrade introduced 3 "new" vulnerabilities.
I was upgrading from version 1.27.0 to the newest one, because the vulnerability with the old Guava version is gone now, but it isn't worth staying up-to-date, when there are 3 new ones.

P.S. Sorry for not filling out the Issue Template, but I don't think, that it is neccessary for just this question.

elharo:
Fixes #618

- [X] Tests pass

This module came in for the first time in 1.28.0for unclear reasons as part of the now reverted split of the apache classes out of the main artifact (#543). I'm proposing removing it as part of the revert of that split.


sduskis:
Wouldn't this be a backwards incompatible change?

elharo:
If we released a new version that changed the signatures that would be incompatible. This is simply not releasing a new version.

chingor13:
Closed in favor of #637

PR #629, opened 2019-04-09
Configurable DNS resolution

george-vacariuc:
Adds the ability to customize the DNS resolution in the Http Client. Provides a sample DNS resolver using the Google DNS resolution over HTTPS.


google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpTransport.java
```diff
@@ -113,7 +116,12 @@ public ApacheHttpTransport() {
    * @since 1.6
    */
   public ApacheHttpTransport(HttpClient httpClient) {
+    this(httpClient, (DnsResolver) null);    
+  }
+
+  public ApacheHttpTransport(HttpClient httpClient, DnsResolver dnsResolver) {
```
chingor13: Please add javadoc for this new constructor.

google-http-client/pom.xml
```diff
@@ -76,6 +76,10 @@
       <groupId>com.google.code.findbugs</groupId>
       <artifactId>jsr305</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.google.code.gson</groupId>
```
chingor13: I'm not seeing where this dependency is needed for this artifact.

google-http-client-dns/src/main/java/com/google/api/client/http/dns/GoogleDnsResolver.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  private ImmutableList<InetAddress> resolveAll(String host) throws Exception {
+    ImmutableList.Builder<InetAddress> result = ImmutableList.<InetAddress>builder();
+    HttpRequestFactory requestFactory = new ApacheHttpTransport().createRequestFactory();
+    HttpRequest request =
+        requestFactory.buildGetRequest(
+            new GenericUrl("https://dns.google.com/resolve?name=" + host));
+    GoogleDnsResponse googleDnsResponse =
+        gson.fromJson(request.execute().parseAsString().toLowerCase(), GoogleDnsResponse.class);
```
chingor13: In general with `google-http-client`, we are JSON implementation agnostic. This means the user selects their JSON parser (gson, jackson2, etc) and provide a `JsonFactory` implementation.

Responses are parsed using `GenericJson` subclasses so the `GoogleDnsResponse` and `GoogleDnsAnswer` classes would extend `GenericJson` and use `@Key` annotations.

google-http-client/src/main/java/com/google/api/client/http/DnsResolver.java
```diff
@@ -0,0 +1,11 @@
+package com.google.api.client.http;
+
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+
+/** A strategy for resolving DNS to IP. */
+public interface DnsResolver {
+
+  /** Resolves a host name to a single IP. */
+  InetAddress resolve(String host) throws UnknownHostException;
```
chingor13: Should this be public?

google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpRequest.java
```diff
@@ -14,11 +14,15 @@
 
 package com.google.api.client.http.apache;
 
+import com.google.api.client.http.DnsResolver;
 import com.google.api.client.http.LowLevelHttpRequest;
 import com.google.api.client.http.LowLevelHttpResponse;
 import com.google.api.client.util.Preconditions;
+import com.google.common.base.Optional;
```
chingor13: This looks unused.

google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpRequest.java
```diff
@@ -60,6 +70,21 @@ public LowLevelHttpResponse execute() throws IOException {
...
+      // Use the default OS DNS resolution.
+      return new ApacheHttpResponse(request, httpClient.execute(request));
+    }
+
+    // Use the IP provided by our specialized DNS resolver.
+    InetAddress ip = dnsResolver.resolve(request.getURI().getHost());
+    int port =
+        request.getURI().getPort() != -1
+            ? request.getURI().getPort()
+            : "https".equals(request.getURI().getScheme()) ? 443 : 80;
```
chingor13: What happens if we provide a URI with weird casing? `URI.create("hTTPs://google.com").getURI().getScheme()` returns `"hTTPs"`. Maybe we don't care what happens in that case.

google-http-client-dns/src/test/java/com/google/api/client/http/dns/ApacheHttpClientWithCustomDnsResolutionTest.java
```diff
@@ -0,0 +1,28 @@
...
+import com.google.api.client.http.GenericUrl;
+import com.google.api.client.http.HttpRequest;
+import com.google.api.client.http.HttpRequestFactory;
+import com.google.api.client.http.apache.ApacheHttpTransport;
+import com.google.api.client.http.dns.GoogleDnsResolver;
+import junit.framework.TestCase;
+
+/** Tests {@link ApacheHttpTransport} using custom DNS resolution. */
+public class ApacheHttpClientWithCustomDnsResolutionTest extends TestCase {
+  public void testConnectWithRoundRobinDnsResolution() throws Exception {
```
chingor13: This looks like an integration test for which we'll have to add credentials.

sduskis:
@george-vacariuc, there's a few TODOs here.  Would you be able to fix them?

PR #635, opened 2019-04-29
Add 404 redirect to new javadoc site

chingor13:
Add a custom 404 page to `gh-pages` which will handle redirects to the new javadocs site.

When the javadocs are published, we can remove the `releases` folder and this 404 page should redirect to the correct page. The node code is used to manually test the `redirectPath` functionality but is not run by a CI suite (doesn't apply to master).

Rather than sharing the same script (and trying to load modules across npm and the browser) the `redirectPath` function is copied from `src/redirects.js` inline in the `404.html`.

### DIFF
--- a/404.html
+++ b/404.html
@@ -1,4 +1,3 @@
-
 <!DOCTYPE html>
 <html>
   <head>
--- a/404.html
+++ b/404.html
@@ -98,7 +98,7 @@
         };
         var destination = redirectUrl(location.href);
         if (!destination) {
-          destination = 'https://googleapis.dev/java/google-cloud-clients/latest/index.html';
+          destination = 'https://googleapis.dev/java/google-http-client/latest/';
         }
         location.href = destination;
       }());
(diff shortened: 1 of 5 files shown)


404.html
```diff
@@ -0,0 +1,107 @@
...
+          var pattern = /^\/google-http-java-client\/releases\/(\d+\.\d+\.\d+|latest)\/javadoc\/(.*)/;
+          var match = pattern.exec(uri.pathname);
+          if (match == null) {
+            return null;
+          }
+          return 'https://googleapis.dev/java/google-http-client/' + match[1] + '/' + match[2] + uri.search + uri.hash;
+        };
+        var destination = redirectUrl(location.href);
+        if (!destination) {
+          destination = 'https://googleapis.dev/java/google-cloud-clients/latest/index.html';
```
JustinBeckwith: Can we rely on the default doc here and just use:
https://googleapis.dev/java/google-cloud-clients/latest/

PR #650, opened 2019-05-24
Fix: remove extraneous content encoding

### LINKED ISSUE #648: HttpContent is encoded two times when posting data
The resulting request is correctly encoded once, but we are encoding an extra time when we calculate the content length.

chingor13:
Fixes #648 

In order to calculate the content length, we must write the encoded the
data first. The encoded data is written to a buffer using a
`ByteArrayOutputStream` implementation and we use that to figure out how
many bytes of data is to be sent.

Previously, this data was thrown away and the content was re-encoded
when it was actually time to send the data.

Instead, we now replace the content with a `ByteArrayContent` which
contains the buffer we wrote to when calculating the size.

We implemented a new `CachingByteArrayOutputStream` so that we can access
the byte buffer directly rather than copying into a new byte array (for
memory performance)


google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -932,7 +932,14 @@ public HttpResponse execute() throws IOException {
         } else {
           contentEncoding = encoding.getName();
           streamingContent = new HttpEncodingStreamingContent(streamingContent, encoding);
-          contentLength = contentRetrySupported ? IOUtils.computeLength(streamingContent) : -1;
+          if (contentRetrySupported) {
+            CachingByteArrayOutputStream outputStream = new CachingByteArrayOutputStream();
```
sduskis: IOUtils checks the size via streaming, and doesn't actually store the data in memory.  The same thing for sending the data.  A large object will currently never fully have to be in memory.

This change may cause an unexpected memory spike for users with large objects

ludoch:
Since this bug was filed on the GAE standard product, memory management is critical as we do not have big mems on instance classes machines...
A test in this env would be welcome (request size limit is 32Mb, compressed or not IIUC).

chingor13:
As is, this implementation won't be good enough

Issue #664, opened 2019-05-31
Cannot parse JSON with big integers


chingor13:
Found this when trying to upgrade the http client in the auth library.

Caused by #529. If the token looks like an integer we now try to parse it as an integer. This number can overflow an integer and make the parser blow up. Additionally, it changed the expected return type which breaks many internal customers.

https://github.com/googleapis/google-auth-library-java/pull/249

TODO:
- [x] revert #529 in #662 
- [x] add tests in #665
- [x] backport to a 1.29 branch and release 1.29.2

PR #665, opened 2019-05-31
Add abstract JSON test and tests for Jackson2 and gson

chingor13:
See #664 

This should fail until we revert #529

### DIFF
--- a/google-http-client-test/src/main/java/com/google/api/client/test/json/AbstractJsonParserTest.java
+++ b/google-http-client-test/src/main/java/com/google/api/client/test/json/AbstractJsonParserTest.java
@@ -1,0 +1,59 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


google-http-client-test/src/main/java/com/google/api/client/test/json/AbstractJsonParserTest.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.JsonObjectParser;
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.math.BigDecimal;
+import java.nio.charset.StandardCharsets;
+import junit.framework.TestCase;
+
+public abstract class AbstractJsonParserTest extends TestCase {
```
sduskis: This repo still uses `junit.framework`?

chingor13: Yeah, it hasn't been a high priority to convert it.

PR #717, opened 2019-06-14
Fix ApacheHttpTransport configuration

### LINKED ISSUE #715: Bug: Apache HttpClient is not properly configured
Just hit this bug.

Versions:
- google-http-client: 1.30.1
- google-http-client-apache-v2: 1.30.1

[Here](https://github.com/googleapis/google-http-java-client/blob/v1.30.1/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpTransport.java#L125-L149), the code is supposed to configure `HttpClient` with some curated values, but in reality, most of them are being ignored. I'll mark below which values are ignored:

```java
    // IGNORED
    SocketConfig socketConfig = SocketConfig.custom()
        .setRcvBufSize(8192)  // IGNORED
        .setSndBufSize(8192)  // IGNORED
        .build();

    PoolingHttpClientConnectionManager connectionManager =
        new PoolingHttpClientConnectionManager(-1, TimeUnit.MILLISECONDS);
    connectionManager.setValidateAfterInactivity(-1);

    return HttpClientBuilder.create()
        .useSystemProperties()
        // IGNORED
        .setSSLSocketFactory(SSLConnectionSocketFactory.getSocketFactory())
        .setDefaultSocketConfig(socketConfig)  // IGNORED
        .setMaxConnTotal(200)  // IGNORED
        .setMaxConnPerRoute(20)  // IGNORED
        .setRoutePlanner(new SystemDefaultRoutePlanner(ProxySelector.getDefault()))
        .setConnectionManager(connectionManager)
        .disableRedirectHandling()
        .disableAutomaticRetries()
```

The reason is apparent from the Javadocs of the methods. For example, `setDefaultSocketConfig` says

```
     * Please note this value can be overridden by the {@link #setConnectionManager(
     *   org.apache.http.conn.HttpClientConnectionManager)} method.
```

That is, calling `setConnectionManager()` makes the custom values set through the methods irrelevant. You can actually check [this behavior](https://github.com/apache/httpcomponents-client/blob/4.5.8/httpclient/src/main/java/org/apache/http/impl/client/HttpClientBuilder.java#L957-L1016) in the `build()` method. For example,

```java
        if (connManagerCopy == null) {
                ...
                // defaultSocketConfig is meaningful only when not calling setConnectionManager().
                poolingmgr.setDefaultSocketConfig(defaultSocketConfig);
```

chingor13:
Fixes #715

Switch back to deprecated `setStaleConnectionCheck`. `HttpClientBuilder` does not expose a way to configure the connection manager's `validateAfterInactivity` configuration. We can either reimplement the entirety of the builder's logic or switch back to a deprecated method.

### DIFF
--- a/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpTransport.java
+++ b/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpTransport.java
@@ -155,19 +155,14 @@ public final class ApacheHttpTransport extends HttpTransport {
                     .setSndBufSize(8192)
                     .build();
 
-    PoolingHttpClientConnectionManager connectionManager =
-            new PoolingHttpClientConnectionManager(-1, TimeUnit.MILLISECONDS);
-    // Disable the stale connection check (previously configured in the HttpConnectionParams
-    connectionManager.setValidateAfterInactivity(-1);
-
     return HttpClientBuilder.create()
             .useSystemProperties()
             .setSSLSocketFactory(SSLConnectionSocketFactory.getSocketFactory())
             .setDefaultSocketConfig(socketConfig)
             .setMaxConnTotal(200)
             .setMaxConnPerRoute(20)
+            .setConnectionTimeToLive(-1, TimeUnit.MILLISECONDS)
             .setRoutePlanner(new SystemDefaultRoutePlanner(ProxySelector.getDefault()))
-            .setConnectionManager(connectionManager)
             .disableRedirectHandling()
             .disableAutomaticRetries();
   }
--- a/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpRequest.java
+++ b/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpRequest.java
@@ -40,7 +40,7 @@ final class ApacheHttpRequest extends LowLevelHttpRequest {
     // disable redirects as google-http-client handles redirects
     this.requestConfig = RequestConfig.custom()
         .setRedirectsEnabled(false)
-        // TODO: configure in HttpClientBuilder when available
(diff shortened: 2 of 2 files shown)


chingor13:
@chanseokoh We will backport this into a 1.30.2 release if it fixes the issue.

google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpTransport.java
```diff
@@ -155,19 +155,14 @@ public static HttpClientBuilder newDefaultHttpClientBuilder() {
...
                     .build();
 
-    PoolingHttpClientConnectionManager connectionManager =
-            new PoolingHttpClientConnectionManager(-1, TimeUnit.MILLISECONDS);
-    // Disable the stale connection check (previously configured in the HttpConnectionParams
-    connectionManager.setValidateAfterInactivity(-1);
-
     return HttpClientBuilder.create()
             .useSystemProperties()
             .setSSLSocketFactory(SSLConnectionSocketFactory.getSocketFactory())
```
chanseokoh: Would you remove this? By default, it will create and use `SSLConnectionSocketFactory.getSocketFactory()`. If this is set, I can't set `.setSSLContext()` or `.setSSLHostnameVerifier()` as they are overridden in the same manner with `setConnectionManager()`.

chanseokoh: Hmm... that may not exactly be same when `useSystemProperties()` is set.

Nonetheless, I think I'm OK either way, since I can explicitly unset the factory by `.setSSLSocketFactory(null)`, so not a big deal.

chanseokoh:
@chingor13 thanks! I'm OK if you make a new release or do a backport. This works in my testing.

PR #738, opened 2019-07-03
Adds CI test for mvn dependency:analyze and fixes missing deps

chingor13:


google-http-client-gson/pom.xml
```diff
@@ -67,11 +67,6 @@
       <artifactId>google-http-client-test</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
```
chingor13: Note: this is inherited from `google-http-client-test` and we inherit from a base test class.

google-http-client-jackson2/pom.xml
```diff
@@ -67,11 +67,6 @@
       <artifactId>google-http-client-test</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
```
chingor13: Note: this is inherited from `google-http-client-test` and we inherit from a base test class.

Issue #740, opened 2019-07-04
HttpResponse should implement Closeable


robertbrignull:
Currently a `HttpResponse` needs to be disconnected to release resources and avoid leaks. This is fine but it can lead to easy mistakes as people new to the library don't realise this or forget.

If `HttpResponse` implemented the `java.io.Closeable` interface then I think there would be a number of benefits:
- Can use try-with-resources. :+1: :+1: :+1: 
- More familiar to developers by implementing the standard interface and using standard method names.
- IDEs and static analysis tools may pick up resource leaks automatically without any extra configuration.

The main downside I can see is that we'd have to keep the `disconnect` method for backwards compatibility and this would mean there would be two interfaces for the same action. Is this a problem? I leave that decision up to you.

I would be happy to implement this if that would help things. It shouldn't be very much work at all. I'm opening this feature request issue in order to gauge if it would be a good idea to implement this or not.

chingor13:
This sounds reasonable and it should be fine to have both for backwards compatibility. We would definitely be willing to accept a PR to add this feature.

PR #744, opened 2019-07-11
make HttpResponse implement Closeable

robertbrignull:
Implements #740.

- [x] Tests pass

---

This PR makes `HttpResponse` implement the `Closeable` interface. I've outlined the benefits that this could give in #740. Hopefully it'll mean
- Can use try-with-resources. :+1: :+1: :+1: 
- More familiar to developers by implementing the standard interface and using standard method names.
- IDEs and static analysis tools may pick up resource leaks automatically without any extra configuration.

However, this does have slightly bigger implications than I originally thought. Consider some code like
```java
Foo foo = request.execute().parseAs(Foo.class);
```
This is a perfectly fine way to use the library, however from a certain perspective it would become invalid if `HttpResponse` were to become closeable because it's "leaking" a closeable object. We know it would still be safe in that it wouldn't leak any resources but it would not be correct from the point of view of the `Closeable` interface because the `close` method is never called.

The danger here is really only that static analyzers might complain, but even that is not too bad as they can normally be silenced or configured to ignore this case. We also don't want to lead people to habitually ignore the `close` method, but that is more a discussion for if `disconnect` is removed or if more implementation is added to `close`.

Note that to be more correct the code could be changed to be like
```java
Foo foo;
try (HttpResponse response = request.execute()) {
    foo = response.parseAs(Foo.class);
}
```
and it would have the same behaviour but is a little bit more verbose.

The other viewpoint is to ignore the above and consider that making `HttpResponse` closeable gives you strictly more options. The above code calling `parseAs` immediately is still correct, however you now also have the option to use try-with-resources to get extra safety when doing something more complicated.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -20,6 +20,7 @@ import com.google.api.client.util.LoggingInputStream;
 import com.google.api.client.util.Preconditions;
 import com.google.api.client.util.StringUtils;
 import java.io.ByteArrayOutputStream;
+import java.io.Closeable;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


chingor13:
Another thing we might need to double check is that the current behavior `disconnect()` is idempotent as `close()` is supposed to be idempotent.

robertbrignull:
> Another thing we might need to double check is that the current behavior `disconnect()` is idempotent as `close()` is supposed to be idempotent.

As far as I can tell they currently all are idempotent. The javadoc doesn't specify this on `HttpRespone` or on `LowLevelHttpResponse` but following through the implementations they do all appear to be safe. Should this be added to the javadoc to make it clearer?

chingor13:
Yeah, let's add to the javadoc for both close and disconnect that the implementation should be idempotent.

PR #749, opened 2019-07-15
Wrap GZIPInputStream with a ConsumingInputStream to drain on close

### LINKED ISSUE #367: chunked encoding + GZIP prevents keep-alive
`GZIPInputStream.read()` starts returning -1 once it finishes reading the gzip trailer. Depending on the underlying `InputStream`, this may occur _before_ it has finished consuming the underlying `InputStream`.

In particular, if the underlying `InputStream` is a `ChunkedInputStream`, the `ChunkedInputStream` never reads the final chunk header (which indicates a chunk of length 0).

When `ChunkedInputStream.close()` is called, it is in a non-STATE_DONE state, and does not call `HttpClient.finished()`, which is required for the `HttpClient` object to get into the keep alive cache.

One possible fix is here:
https://github.com/google/google-http-java-client/blob/acdaeacc3c43402d795266d2ebcb584abe918961/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java#L363

Rather than returning a `GZIPInputStream`, return a proxy input stream that retains references to both the gzip stream and the underlying stream. The proxy's `close()` method first consumes the underlying stream by calling `read()` and then calls `close()` on the gzip stream.

chingor13:
Fixes #367

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -331,7 +331,8 @@ public final class HttpResponse {
           if (!returnRawInputStream
               && contentEncoding != null
               && contentEncoding.contains("gzip")) {
-            lowLevelResponseContent = new GZIPInputStream(lowLevelResponseContent);
+            lowLevelResponseContent = new ConsumingInputStream(
+                new GZIPInputStream(lowLevelResponseContent));
           }
           // logging (wrap content with LoggingInputStream)
           Logger logger = HttpTransport.LOGGER;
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -356,6 +357,44 @@ public InputStream getContent() throws IOException {
     return content;
   }
 
+  static class ConsumingInputStream extends InputStream {
```
elharo: An outer class would better separate unrelated code. This is complex enough to justify it.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -356,6 +357,44 @@ public InputStream getContent() throws IOException {
...
+    @Override
+    public int read(byte[] b, int off, int len) throws IOException {
+      return inputStream.read(b, off, len);
+    }
+
+    @Override
+    public void close() throws IOException {
+      if (!closed && inputStream != null) {
+        try {
+          drainInputStream(this);
```
elharo: Why is it important to drain the input stream and throw away the excess data, instead of simply closing the stream?

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
```diff
@@ -457,4 +459,40 @@ public LowLevelHttpResponse execute() throws IOException {
...
         request.execute().getContent() instanceof GZIPInputStream);
   }
+
+  public void testGetContent_gzipEncoding_finishReading() throws IOException {
+    byte[] dataToCompress = "abcd".getBytes(StandardCharsets.UTF_8);
+    byte[] mockBytes;
+    try (ByteArrayOutputStream byteStream = new ByteArrayOutputStream(dataToCompress.length)) {
+      GZIPOutputStream zipStream = new GZIPOutputStream((byteStream));
+      zipStream.write(dataToCompress);
+      zipStream.close();
```
elharo: We should probably also have a case where there's extra data on the stream after the gzipped content, since handling that is the purpose of this PR.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -356,6 +357,44 @@ public InputStream getContent() throws IOException {
...
+          inputStream.close();
+        } finally {
+          this.closed = true;
+        }
+      }
+    }
+
+    static void drainInputStream(final InputStream inputStream) throws IOException {
```
elharo: I think this method can be private. Also you've reinvented ByteStreams.exhaust from Guava.

PR #752, opened 2019-07-25
Add support for nanosecond precision when parsing rfc3339 strings

olavloite:
The original parseRfc3339 method of `DateTime` returns a `DateTime` instance that has millisecond precision. The parse method therefore also disregards any time information beyond millisecond precision. For Spanner we would like to add nanosecond precision, and instead of replicating the parse method in the Spanner specific libraries, this change will add nanosecond support to this parse method, while retaining the original behavior of the parseRfc3339 method.

The original parseRfc3339 method allows strings to contain more than millisecond precision. It does however not specify exactly what it does with any additional information, and the behavior is not consistent:
* The input value `2018-12-31T23:59:59.9999Z` would be truncated to `2018-12-31T23:59:59.999Z`
* The input value `2018-12-31T23:59:59.999999999Z` would be rounded up to `2019-01-01T00:00:00Z`

After this change the method will always truncate any time information beyond millisecond precision when returning a `DateTime` instance.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/DateTime.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/DateTime.java
@@ -14,12 +14,15 @@
 
 package com.google.api.client.util;
 
+import com.google.common.base.Strings;
 import java.io.Serializable;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.GregorianCalendar;
+import java.util.Objects;
 import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -39,12 +42,12 @@ public final class DateTime implements Serializable {
   private static final TimeZone GMT = TimeZone.getTimeZone("GMT");
 
   /** Regular expression for parsing RFC3339 date/times. */
-  private static final Pattern RFC3339_PATTERN =
-      Pattern.compile(
-          "^(\\d{4})-(\\d{2})-(\\d{2})" // yyyy-MM-dd
-              + "([Tt](\\d{2}):(\\d{2}):(\\d{2})(\\.\\d+)?)?" // 'T'HH:mm:ss.milliseconds
-              + "([Zz]|([+-])(\\d{2}):(\\d{2}))?"); // 'Z' or time zone shift HH:mm following '+' or
-                                                    // '-'
+  private static final String RFC3339_REGEX =
+      "(\\d{4})-(\\d{2})-(\\d{2})" // yyyy-MM-dd
+          + "([Tt](\\d{2}):(\\d{2}):(\\d{2})(\\.\\d{1,9})?)?" // 'T'HH:mm:ss.nanoseconds
+          + "([Zz]|([+-])(\\d{2}):(\\d{2}))?"; // 'Z' or time zone shift HH:mm following '+' or '-'
+
+  private static final Pattern RFC3339_PATTERN = Pattern.compile(RFC3339_REGEX);
 
   /**
    * Date/time value expressed as the number of ms since the Unix epoch.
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/util/DateTime.java
```diff
@@ -260,6 +263,11 @@ public int hashCode() {
    * NumberFormatException}. Also, in accordance with the RFC3339 standard, any number of
    * milliseconds digits is now allowed.
    *
+   * <p>Any time information beyond millisecond precision will be truncated. Prior to version 1.30.2
```
elharo: will be --> is (tech writing lives in the eternal present)

google-http-client/src/main/java/com/google/api/client/util/DateTime.java
```diff
@@ -260,6 +263,11 @@ public int hashCode() {
...
+   * this method did not have a well-defined behavior of what would happen with any time information
```
elharo: No need to document what happened in earlier versions here.

google-http-client/src/main/java/com/google/api/client/util/DateTime.java
```diff
@@ -269,6 +277,98 @@ public int hashCode() {
...
    */
   public static DateTime parseRfc3339(String str) throws NumberFormatException {
+    return parseRfc3339WithNanoSeconds(str).toDateTime();
+  }
+
+  /**
+   * Parses an RFC3339 timestamp to a pair of seconds and nanoseconds since Unix Epoch.
+   *
+   * @param str Date/time string in RFC3339 format
+   * @throws NumberFormatException if {@code str} doesn't match the RFC3339 standard format; an
```
elharo: This does not seem like the right exception type. IllegalArgumentException would be better.

google-http-client/src/main/java/com/google/api/client/util/DateTime.java
olavloite: Changed.

google-http-client/src/main/java/com/google/api/client/util/DateTime.java
olavloite: Removed unnecessary documentation.

google-http-client/src/main/java/com/google/api/client/util/DateTime.java
olavloite: I've changed the declared exception type.

olavloite:
> No comment on the design yet, but adding public API normally suggests incrementing the minor version. In this case, I think it would need to bump to 1.4.0 in the pom.xml.

Shouldn't that be 1.31.0?

elharo:
You're right. It should be 1.31.0 I get confused when version numbers hit double digits. :-)

google-http-client/src/main/java/com/google/api/client/util/DateTime.java
```diff
@@ -269,6 +274,98 @@ public int hashCode() {
...
+   *     exception is thrown if {@code str} doesn't match {@code RFC3339_REGEX} or if it contains a
+   *     time zone shift but no time.
+   */
+  public static SecondsAndNanos parseRfc3339ToSecondsAndNanos(String str)
+      throws IllegalArgumentException {
+    return parseRfc3339WithNanoSeconds(str).toSecondsAndNanos();
+  }
+
+  /** A timestamp represented as the number of seconds and nanoseconds since Epoch. */
+  public static final class SecondsAndNanos implements Serializable {
```
elharo: I'm not sure what our plans are for Java 8, but if we could use java.time I think we could avoid introducing this class and associated public API. @chingor13 has any final decision been made on that?

chingor13: This library is very heavily used on android, so even if we move google-cloud-java to Java 8, we'd likely avoid using new features for a long time. java.time was introduced in Android at API level 26 (Oreo).

google-http-client/src/test/java/com/google/api/client/util/DateTimeTest.java
```diff
@@ -142,6 +146,89 @@ public void testParseRfc3339() {
...
+   * 2018-10-28T03:00:00Z                |   1540695600  |   0
+   * 2018-10-28T02:30:00Z                |   1540693800  |   0
+   * 2018-03-01T10:11:12.123Z            |   1519899072  |   123000000
+   * 2018-10-28T02:30:00+02:00           |   1540686600  |   0
+   * 2018-03-01T10:11:12.123456789Z      |   1519899072  |   123456789
+   * 2018-03-01T10:11:12.1000Z           |   1519899072  |   100000000
+   * </pre>
+   */
+  public void testParseRfc3339ToSecondsAndNanos() {
+    Map<String, SecondsAndNanos> map = new HashMap<>();
```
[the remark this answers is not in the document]
olavloite: That's a good point. I've changed the test case accordingly.

PR #757, opened 2019-08-03
Surface rawToken for JWT wrapper

salrashid123:
Adds ability to recall the raw JWT encoded token from `JsonWebSignature` and `JsonWebToken`.

The current .`toString()` method for both classes returns a JSON object.  Its useful to have access to the raw token directly to be used in an http client (as `Authorization: Bearer <rawToken>`, for example)

ref:
  https://github.com/googleapis/google-auth-library-java/pull/303#pullrequestreview-269723158

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
+++ b/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
@@ -482,6 +482,11 @@ public class JsonWebSignature extends JsonWebToken {
     return signedContentBytes;
   }
 
+  /** Returns the URLsafe base64 encoded JWS token */
+  public String getRawToken() {
+    return super.getRawToken() + "." + Base64.encodeBase64URLSafeString(getSignatureBytes());
+  }
+
   /**
    * Parses the given JWS token string and returns the parsed {@link JsonWebSignature}.
    *
--- a/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
+++ b/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
@@ -482,7 +482,7 @@ public class JsonWebSignature extends JsonWebToken {
     return signedContentBytes;
   }
 
-  /** Returns the URLsafe base64 encoded JWS token */
+  /** Returns the URL safe Base64 encoded JWS token. */
   public String getRawToken() {
     return super.getRawToken() + "." + Base64.encodeBase64URLSafeString(getSignatureBytes());
   }
--- a/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebToken.java
+++ b/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebToken.java
@@ -15,6 +15,7 @@
 package com.google.api.client.json.webtoken;
 
 import com.google.api.client.json.GenericJson;
+import com.google.api.client.util.Base64;
 import com.google.api.client.util.Key;
 import com.google.api.client.util.Objects;
 import com.google.api.client.util.Preconditions;
@@ -359,6 +360,12 @@ public class JsonWebToken {
   @Override
   public String toString() {
     return Objects.toStringHelper(this).add("header", header).add("payload", payload).toString();
+  }
+
+  public String getRawToken() {
+    return Base64.encodeBase64URLSafeString(header.toString().getBytes())
+        + "."
+        + Base64.encodeBase64URLSafeString(payload.toString().getBytes());
   }
 
   /**
--- a/google-http-client/src/test/java/com/google/api/client/json/webtoken/JsonWebSignatureTest.java
+++ b/google-http-client/src/test/java/com/google/api/client/json/webtoken/JsonWebSignatureTest.java
@@ -49,8 +49,8 @@ public class JsonWebSignatureTest extends TestCase {
   public void testRawToken() throws Exception {
     String rawToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6Ikp"
         + "vaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
-    JsonWebSignature jws = JsonWebSignature.parse(new MockJsonFactory(), rawToken);
-    assertEquals(jws.getRawToken(), rawToken);
+    JsonWebSignature signature = JsonWebSignature.parse(new MockJsonFactory(), rawToken);
+    assertEquals(signature.getRawToken(), rawToken);
   }
 
   private X509Certificate verifyX509WithCaCert(TestCertificates.CertData caCert) throws Exception {
--- a/google-http-client/src/test/java/com/google/api/client/json/webtoken/JsonWebSignatureTest.java
+++ b/google-http-client/src/test/java/com/google/api/client/json/webtoken/JsonWebSignatureTest.java
@@ -46,6 +46,13 @@ public class JsonWebSignatureTest extends TestCase {
         JsonWebSignature.signUsingRsaSha256(privateKey, new MockJsonFactory(), header, payload));
   }
 
+  public void testRawToken() throws Exception {
+    String rawToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6Ikp"
+        + "vaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
+    JsonWebSignature jws = JsonWebSignature.parse(new MockJsonFactory(), rawToken);
+    assertEquals(jws.getRawToken(), rawToken);
+  }
+
   private X509Certificate verifyX509WithCaCert(TestCertificates.CertData caCert) throws Exception {
     JsonWebSignature signature = TestCertificates.getJsonWebSignature();
     X509TrustManager trustManager = caCert.getTrustManager();


google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
```diff
@@ -482,6 +482,11 @@ private static X509TrustManager getDefaultX509TrustManager() {
     return signedContentBytes;
   }
 
+  /** Returns the URLsafe base64 encoded JWS token */
```
elharo: "URL safe, Base64"

period at end

google-http-client/src/test/java/com/google/api/client/json/webtoken/JsonWebSignatureTest.java
```diff
@@ -46,6 +46,13 @@ public void testSign() throws Exception {
         JsonWebSignature.signUsingRsaSha256(privateKey, new MockJsonFactory(), header, payload));
   }
 
+  public void testRawToken() throws Exception {
+    String rawToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6Ikp"
+        + "vaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
+    JsonWebSignature jws = JsonWebSignature.parse(new MockJsonFactory(), rawToken);
```
elharo: jws --> signature

google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebToken.java
```diff
@@ -361,6 +362,12 @@ public String toString() {
     return Objects.toStringHelper(this).add("header", header).add("payload", payload).toString();
   }
 
+  public String getRawToken() {
+    return Base64.encodeBase64URLSafeString(header.toString().getBytes())
+        + "."
```
elharo: Is there supposed to be a signature here too? Or can there be?

salrashid123: nope, the `JsonWebToken` class doens't have a place for the signature (it only accepts the `Heaer` and `Payload`; the `JsonWebSignature` one extends the base to tack on the signature bit

elharo: I'm trying to work my way through RFC 7519. Is it possible for there to be a JWT that is neither a JWE or a JWS? Section seems to suggest that this is possible, but that the signature in this case is the empty string and there should be a terminating period.

salrashid123: oh, you mean if `JsonWebToken.getRawToken()` should emit a `foo.bar.`  and the only thinkg `JsonWebSignature.getRawToken()` does is direct appending (w/o a '.' prefix)?  good question, i don't know that much detail in the specs...(IMO i doubt it but i gotta do some reading tomorrwo)

salrashid123: ok, AFAIK, the trailing `.` isn't needed at all in the output of `JsonWebToken.class` ..its just a delimiter...  infact, i'm not sure where i'd use just the output of `JsonWebToken.getRawToken` standalone ouside of this context ..i'm guessing that was added incase we add on a `JsonWebEncryption`... i'd only be interested in `JWS`  or maybe `JWE`

google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebToken.java
```diff
@@ -361,6 +362,12 @@ public String toString() {
     return Objects.toStringHelper(this).add("header", header).add("payload", payload).toString();
   }
 
+  public String getRawToken() {
```
elharo: This method needs JavaDoc. Also note that the word "raw" does not appear in the RFC. Is there a term in the RFC that does describe this?

elharo: Reading the spec, I think what this is, is a "compact serialized form". Furthermore, I'm beginning to think it belongs in the JsonWebSignature class rather than JsonWebToken, because it isn't defined for JsonWebToken.

google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
```diff
@@ -482,6 +482,11 @@ private static X509TrustManager getDefaultX509TrustManager() {
     return signedContentBytes;
   }
 
+  /** Returns the URL safe Base64 encoded JWS token. */
+  public String getRawToken() {
```
elharo: Since we're unlikely to get the API perfect on the first try, we should mark this @Beta to warn users it might change.

google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebToken.java
elharo: Since we're unlikely to get the API perfect on the first try, we should mark this @Beta to warn users it might change.

google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
elharo: `@Override`

google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebToken.java
salrashid123: cloest term maybe `getEncodedToken()`

salrashid123: yeah, thats kind of what i meant that i don't know where i'd use the internal method directly (i.,e `JsonWebToken.getRawToken()`.  i'lonly make this visible to `JsonWebToken` class

salrashid123:
ok, this became a bit more complicated.  The classes are structured such that the original `tokenString`  in `JsonWebSignature parse(String tokenString)`  gets marshalled into a json object which itself sorts the claims.  the reverse mechanism i use to unmarshall creates a different string representaton (i.,e the order is different).

so a snippet like this

```java
String rawToken = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6Ikp"
+ "vaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c";
JsonWebSignature signature = JsonWebSignature.parse(JacksonFactory.getDefaultInstance(), rawToken);
System.out.println(signature.getEncodedToken());
System.out.println(rawToken);
```

gives 
```json
{
  "alg": "HS256",
  "typ": "JWT"
}.
{
  "iat": 1516239022,
  "sub": "1234567890",
  "name": "John Doe"
}
```

vs

```json
{
  "alg": "HS256",
  "typ": "JWT"
}.
{
  "sub": "1234567890",
  "name": "John Doe",
  "iat": 1516239022
}
```

elharo:
Per the [JSON specification]( http://www.json.org/) "An object is an unordered set of name/value pairs". Nothing here, tests inlcuded, should depend on this order.

salrashid123:
The ordering thing doesn't matter for JSON but in this case i'm 'attaching' a singature part of some claims that were in a specifc order to another..the net result is the JWT string representation that the output of these methods isn't valid anymore.

For example, if i take a valid `rawToken` that i know works on a serivce like Cloud Run

```
String rawToken = "a.b.c";
JsonWebSignature signature = JsonWebSignature.parse(JacksonFactory.getDefaultInstance(), rawToken);
System.out.println(signature.getEncodedToken());
```

the output of `signatuere.getEncodedToken()` isn't the same jwt and doens't work anymore against Cloud Run.  Basically, i probably should figure out how to 'just save and return' the raw token thats passed into the `.parse()` object w/o marshalling/unmarshalling....

elharo:
The signature should be independent of order as well. If it isn't, then there's a bug somewhere, maybe more than one. Are you saying that

```
{
  "iat": 1516239022,
  "sub": "1234567890",
  "name": "John Doe"
}
```

produces a different signature than 

```
{
  "sub": "1234567890",
  "name": "John Doe"
  "iat": 1516239022,
}
```

salrashid123:
the order does matter...to sign, basically i'll take

```
{
  "iat": 1516239022,
  "sub": "1234567890",
  "name": "John Doe"
}
```
use it as string, covert to base64 then digitally sign that to get the signature

which isn't going to produce the same out put if i alter the claims around.

elharo:
I need to check the spec, but if that's really how the JWT spec defines signing, then the spec is broken. The usual way of handling this problem is defininf a canonical and reprodubile ordering; e.g. like canonical XML does for attributes.

elharo:
Also, I think we might be jumping too fast to code here. This is more complex than I realized and I'm not sure we understand the problem sufficiently well. It's probabaly worth starting with a design doc that lays out the plansd, tradeoffs, and issues involed.

elharo:
So I've confirmed that the spec is indeed broken. You will need to figure out how to 'just save and return' the raw token that's passed into the .parse() object w/o marshalling/unmarshalling....

salrashid123:
yep...thats what i figured (i initially started this thinking the 'easiest' way was to 'just save and retun' the token thats passed in thorugh this

```java 
public JsonWebSignature parse(String tokenString)
```

...but then realized there is also a constructor that a constructor which immediately messes up the ordering

```java
public JsonWebSignature(
      Header header, Payload payload, byte[] signatureBytes, byte[] signedContentBytes) {
```

what that means is while i could use the token passed into `parse()` anyone using the direct constructor won't be able to get `.getEncodedToken()` properly.

I could just throw an exception if someone tries to use that method without going through static `.parse()`....but that doens't seem all too clean...

elharo:
Can you remove that constructor or make it private?

salrashid123:
i don't think i can remove or make it not visible (it'd break anyone using it (including our tests):
  https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/testing/json/webtoken/TestCertificates.java#L312

i could create second constructor pass through the string format  (one that'd set the raw token and then call the existing constructor.

```java
public JsonWebSignature(
      Header header, Payload payload, byte[] signatureBytes, byte[] signedContentBytes, String encodedToken)
```

at this point, i kinda don't know what to do since there doesn't seem to be a good way out of this

elharo:
You could deprecate this constructor. Might be worth doing that if it's definitely broken like this.

elharo:
I think I see a way to fix this. We need to canonicalize field order before signing, and do all our own signing. It will need some new runtime exceptions, but should be plausible. I do wonder if it might be easier to build a better library. This class has other problems too, such as exposing mutable internal state.

elharo:
@benwhitehead suggested we look at https://github.com/auth0/java-jwt to see if it might work better than rolling our own.

salrashid123:
ugh...this is a bit of a mess..I'm not sure if canonicalizing before signing is even an option anymore...
the existing  surface
```java
public JsonWebSignature(
      Header header, Payload payload, 
      byte[] signatureBytes, 
      byte[] signedContentBytes, 
```
precludes us from doing anything  since the `Payload` is already provided in whatever alpha(?) order it uses...and the signed portion is separate (meaning it was signed already with whatever propr string format it was...).  

This means JWT Header, Payload objects used in the code is disconnected from the signature that was itself passed

better way maybe to just do a rewrite after annotating as deprecated and use the auth0 stuff

elharo:
I think it's legitimate to have the `public JsonWebSignature(Header header, Payload payload, byte[] signatureBytes, byte[] signedContentBytes)` throw a runtime exception if the signatureBytes and signedContentBytes don't match the header and payload. It's API compatible, and a runtime exception works because this is an illegal argument problem, i.e. a programming error on the part of the caller. 

If we use auth0 perhaps that could be an internal implementation detail?

PR #783, opened 2019-08-08
add module name to submodules

elharo:
@chingor13


elharo:
done. @suztomo is working on a general check to find all the various artifacts across the GCP orbit that need this treatment. Foe some of them I think we'll have to fix split packages first.

elharo:
also see https://github.com/GoogleCloudPlatform/cloud-opensource-java/pull/835

PR #784, opened 2019-08-08
Switch to google_checks.xml for maven-checkstyle-plugin

### LINKED ISSUE #781: Update checkstyle.xml
Current google checkstyle config is at 

https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml

chingor13:
Fixes #781


chingor13:
Some of these rules conflict with our code formatter. We'll want to add more rules to our suppression list.

elharo:
There might be some issues with the latest released google_checks.xml not being quite up-to-date with google style. I seem to recall hitting this in the past.

samples/dailymotion-simple-cmdline-sample/pom.xml
```diff
@@ -37,7 +37,7 @@
         <artifactId>maven-checkstyle-plugin</artifactId>
         <version>3.0.0</version>
```
elharo: You can upgrade this to 3.1.0. That might have slightly more up to date google checks.

chingor13: I'll try that, but will have to only run it on Java 8+

elharo: In that case I wonder if we could simply copy and paste the newer googlechecks.xml into our own checkstyle config and stay at 3.0.0?

chingor13:
Alternative to pull in google_checks.xml as our base checkstyle.xml in #786

chingor13:
closed in favor of #786

PR #803, opened 2019-08-28
fix: correct connect timeout setting for ApacheHttpRequest

### LINKED ISSUE #799: Regression: connection timeout ignored in 1.31.0 while works in 1.27.0
#### Environment details

The same environment as in #795.

#### Steps to reproduce

  1. `.setConnectTimeout(4000)`
  2. Call `https://example.com:81`. It always times out after 30 seconds (on my machine) with 1.31.0. See the example code below.
     - `example.com` is reserved by the DNS standard, so this will always trigger timeout.

Note that I haven't tested read timeout.

#### Code example

* 1.31.0: output proves that it times out after 30 seconds.
```java
public class Main {

  public static void main(String[] args) throws IOException {
    GenericUrl url = new GenericUrl("https://example.com:81");
    HttpRequest req = new ApacheHttpTransport().createRequestFactory().buildGetRequest(url).setConnectTimeout(4000);
    long started = System.nanoTime();
    try {
      req.execute();
    } finally {
      System.out.println("elapsed (s): " + (System.nanoTime() - started) / 1000000000L);
    }
  }
}
```

```
elapsed (s): 30
Exception in thread "main" org.apache.http.conn.HttpHostConnectException: Connect to example.com:81 [example.com/93.184.216.34, example.com/2606:2800:220:1:248:1893:25c8:1946] failed: Connection timed out (Connection timed out)
	at org.apache.http.impl.conn.DefaultHttpClientConnectionOperator.connect(DefaultHttpClientConnectionOperator.java:156)
```

* 1.27.0: connect timeout is honored.

```
elapsed (s): 4
Exception in thread "main" java.net.SocketTimeoutException: connect timed out
	at java.net.PlainSocketImpl.socketConnect(Native Method)
```

#### Any additional information below

I noticed another behavioral change from 1.27.0 to 1.31.0 that 1.31.0 throws `HttpHostConnectException` while 1.27.0 throws `java.net.SocketTimeoutException`, but I think it is a bug in the Apache library in `DefaultHttpClientConnectionOperator`. I will file a bug against them.

```java
            } catch (final ConnectException ex) {
                if (last) {
                    final String msg = ex.getMessage();
                    throw "Connection timed out".equals(msg)
                                    ? new ConnectTimeoutException(ex, host, addresses)
                                    : new HttpHostConnectException(ex, host, addresses);
                }
```

chingor13:
Fixes #799

### DIFF
--- a/google-http-client-apache-v2/src/test/java/com/google/api/client/http/apache/v2/ApacheHttpTransportTest.java
+++ b/google-http-client-apache-v2/src/test/java/com/google/api/client/http/apache/v2/ApacheHttpTransportTest.java
@@ -190,7 +190,7 @@ public class ApacheHttpTransportTest {
     HttpTransport httpTransport = new ApacheHttpTransport();
     GenericUrl url = new GenericUrl("http://google.com:81");
     try {
-      httpTransport.createRequestFactory().buildGetRequest(url).setConnectTimeout(5).execute();
+      httpTransport.createRequestFactory().buildGetRequest(url).setConnectTimeout(100).execute();
       fail("should have thrown an exception");
     } catch (HttpHostConnectException expected) {
       // expected
--- a/google-http-client-apache-v2/src/test/java/com/google/api/client/http/apache/v2/ApacheHttpTransportTest.java
+++ b/google-http-client-apache-v2/src/test/java/com/google/api/client/http/apache/v2/ApacheHttpTransportTest.java
@@ -18,6 +18,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.junit.Assume.assumeTrue;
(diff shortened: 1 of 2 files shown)


chingor13:
I'm not sure why this fails on windows, but it seems to be within the Apache adapter. We should be able to skip this test on windows.

google-http-client-apache-v2/src/test/java/com/google/api/client/http/apache/v2/ApacheHttpTransportTest.java
```diff
@@ -183,6 +187,23 @@ public void process(HttpRequest request, HttpContext context)
     assertTrue("Expected to have called our test interceptor", interceptorCalled.get());
   }
 
+  @Test(timeout = 10_000L)
+  public void testConnectTimeout() {
+    // Apache HttpClient doesn't appear to behave correctly on windows
```
elharo: Is there a bug filed against Apache we could reference here?

PR #807, opened 2019-08-30
docs: migrate docs into source control from the wiki

chingor13:
Once moved here, we can disable the wiki and enable GitHub pages from the root docs/ folder on master


chingor13:
Note that I arbitrarily picket the "dinky" built-in template because it looked clean and had a sidebar location for putting navigation"

Issue #621, opened 2019-03-25
shouldn't have a hard-dependency on a <1.0 api


adriancole:
Even when not using tracing, opencensus 0.18 types are required per this line in `HttpRequest`

```
  /** OpenCensus tracing component. */
private final Tracer tracer = OpenCensusUtils.getTracer();
```

grpc code is more defensive to not require loading classes until used.

elharo:
Relevant: https://github.com/GoogleCloudPlatform/cloud-opensource-java/blob/master/library-best-practices/JLBP-4.md

adriancole:
why is this a question and not a bug?

bogdandrutu:
OpenCensus versions <1.0 should be treated as any version >1.0. In the OpenCensus project we made a decision to respect all the requirements for versions >1.0 which means we do not break compatibility (remove classes/methods or change signatures).

So this should be closed.

elharo:
If that's so, you might as well go ahead and release 1.0. However I have to note that for an allegedly stable product, the different version of OpenCensus have caused us a lot of trouble.

adriancole:
Even if there is no argument about version drift etc, it is still completely unnecessary for this code to be called in absence of any setup instructions. 1.0 or not, it is pulling unnecessary dependencies into the application, and for no reason as the tracer isn't used.

adriancole:
the typical way to fix accidental classpath leak is to make a separate class that has the stuff inside. There are varying approaches to this. For example armeria recently made it possible to not have dropwizard in the classpath via this change https://github.com/line/armeria/pull/2107

At some point, there might be some RPC library that you also find troublesomely hard pinned to something you don't want or use. I hope pointers on technical solution is helpful.

adriancole:
thanks, please note which version releases this so we can align deps!

adriancole:
PS It looks like the change was to remove OpenCensus. If that's something that needs to be reverted, I'd recommend having a look at the comment I mentioned here https://github.com/googleapis/google-http-java-client/issues/621#issuecomment-536355409

Basically, there are ways to perform the change in such a way that the census code still works, but is itself optional. If it isn't clear, and someone says "pull requests welcome" I would rather help implement that vs reverting back to having a strict dep on census again.

codyoss:
I will leave this open, but this has been deprioritized. Down the road I think we might move towards using open telemetry. Until this, this will sit.

elharo:
[...]

PR #837, opened 2019-10-11
revert: "feat: remove opencensus (#834)"

chingor13:
#621


chingor13:
We can make it an optional dependency before the opentelemetry conversion. So rather than ripping it out and then reimplementing, we can make it defensive in a single PR.

PR #840, opened 2019-10-14
feat: wrap GZIPInputStream for connection reuse

codyoss:
If a connection is closed and there are some bytes that have not
been read that connection can't be reused. Now GZIPInputStream
will have all of its bytes read on close automatically to promote
connection reuse.

Cherry-picked: #749
Fixes: #367

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.api.client.http;
+
+import com.google.api.client.util.Preconditions;
+import com.google.common.io.ByteStreams;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This class in meant to wrap an {@link InputStream} so that all bytes in the steam are read and
+ * discarded on {@link InputStream#close()}. This ensures that the underlying connection has the
+ * option to be reused.
+ */
+final class ConsumingInputStream extends InputStream {
+  private InputStream inputStream;
+  private boolean closed = false;
+
+  ConsumingInputStream(InputStream inputStream) {
+    this.inputStream = Preconditions.checkNotNull(inputStream);
+  }
+
+  @Override
+  public int read() throws IOException {
+    return inputStream.read();
+  }
+
+  @Override
+  public int read(byte[] b, int off, int len) throws IOException {
+    return inputStream.read(b, off, len);
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (!closed && inputStream != null) {
+      try {
+        ByteStreams.exhaust(this);
+        inputStream.close();
+      } finally {
+        this.closed = true;
+      }
+    }
+  }
+}
--- a/google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
@@ -16,8 +16,8 @@
 
 package com.google.api.client.http;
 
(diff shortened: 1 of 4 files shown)


codyoss:
Noticed the #749 had not been updated in a while so I cherry picked and fixed the PR comments.

google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Override
+  public int read(byte[] b, int off, int len) throws IOException {
+    return inputStream.read(b, off, len);
+  }
+
+  @Override
+  public void close() throws IOException {
+    if (!closed && inputStream != null) {
```
elharo: I don't think inputStream can be null here.

google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
```diff
@@ -0,0 +1,58 @@
...
+      try {
+        ByteStreams.exhaust(this);
```
elharo: I wonder if we could have an infinite loop here. See internal discussion.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -331,7 +331,8 @@ public InputStream getContent() throws IOException {
           if (!returnRawInputStream
               && contentEncoding != null
               && contentEncoding.contains("gzip")) {
```
elharo: not a new bug, but "contains" is incorrect. This should be equals or perhaps equalsIgnoreCase with a trim, or something like that. However it shouldn't fire on foozipbar.

filed https://github.com/googleapis/google-http-java-client/issues/842

google-http-client/src/test/java/com/google/api/client/http/ConsumingInputStreamTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import org.junit.Test;
+
+public class ConsumingInputStreamTest {
+
+  @Test
+  public void testClose_drainsBytesOnClose() throws IOException {
+    MockInputStream mockInputStream = new MockInputStream("abc123".getBytes());
```
elharo: antipattern: using getBytes without character encoding

google-http-client/src/test/java/com/google/api/client/http/ConsumingInputStreamTest.java
```diff
@@ -0,0 +1,63 @@
...
+    InputStream consumingInputStream = new ConsumingInputStream(mockInputStream);
+
+    assertEquals(6, mockInputStream.getBytesToRead());
```
elharo: this would fail if the default system encoding is UTF-16, for example

google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
```diff
@@ -0,0 +1,58 @@
...
+import com.google.common.io.ByteStreams;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This class in meant to wrap an {@link InputStream} so that all bytes in the steam are read and
+ * discarded on {@link InputStream#close()}. This ensures that the underlying connection has the
+ * option to be reused.
+ */
+final class ConsumingInputStream extends InputStream {
```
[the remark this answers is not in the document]
codyoss: sure

google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
codyoss: nope, it can now though.

google-http-client/src/test/java/com/google/api/client/http/ConsumingInputStreamTest.java
codyoss: Whoops! Good catch.

google-http-client/src/main/java/com/google/api/client/http/ConsumingInputStream.java
```diff
@@ -0,0 +1,58 @@
...
+package com.google.api.client.http;
+
+import com.google.api.client.util.Preconditions;
+import com.google.common.io.ByteStreams;
+import java.io.IOException;
+import java.io.InputStream;
+
+/**
+ * This class in meant to wrap an {@link InputStream} so that all bytes in the steam are read and
+ * discarded on {@link InputStream#close()}. This ensures that the underlying connection has the
```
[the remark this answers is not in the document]
codyoss: That is a very good question. I am not sure.

codyoss:
@elharo I fixed all of the comments. Do you still have reservations around a possible infinite stream? If so, any thought on how we should move forward with this?(Maybe we just don't if we think the infinite stream is too big of a risk as this feature requires exhausting all of the extra bytes for it to work.)

PR #843, opened 2019-10-14
fix: HttpResponse GZip content encoding equality change

### LINKED ISSUE #842: HttpResponse uses contains where it needs equals
```
         if (!returnRawInputStream
              && contentEncoding != null	              && contentEncoding != null
              && contentEncoding.contains("gzip")) {
```

"contains" is incorrect. This should be equals or perhaps equalsIgnoreCase with a trim, or something like that. However it shouldn't fire on foogzipbar.

egsavage:
Fixes #842 by having HttpResponse compare (case insensitive) the content-encoding against "gzip" and "x-gzip" (HTTP 1.1).

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -80,6 +80,12 @@ public final class HttpResponse {
   /** Whether {@link #getContent()} should return raw input stream. */
   private final boolean returnRawInputStream;
 
+  /** Content encoding for GZip */
+  private static final String CONTENT_ENCODING_GZIP = "gzip";
+
+  /** Content encoding for GZip HTTP 1.1 */
+  private static final String CONTENT_ENCODING_XGZIP = "x-gzip";
+
   /**
    * Determines the limit to the content size that will be logged during {@link #getContent()}.
    *
@@ -330,7 +336,8 @@ public final class HttpResponse {
           String contentEncoding = this.contentEncoding;
           if (!returnRawInputStream
               && contentEncoding != null
-              && contentEncoding.contains("gzip")) {
+              && (CONTENT_ENCODING_GZIP.equalsIgnoreCase(contentEncoding.trim())
+                  || CONTENT_ENCODING_XGZIP.equalsIgnoreCase(contentEncoding.trim()))) {
             lowLevelResponseContent = new GZIPInputStream(lowLevelResponseContent);
           }
           // logging (wrap content with LoggingInputStream)
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -335,8 +335,8 @@ public final class HttpResponse {
         try {
           // gzip encoding (wrap content with GZipInputStream)
           if (!returnRawInputStream && this.contentEncoding != null) {
-            final String oontentEncoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
-            if (CONTENT_ENCODING_GZIP.equals(oontentEncoding) || CONTENT_ENCODING_XGZIP.equals(oontentEncoding)) {
+            String oontentencoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
+            if (CONTENT_ENCODING_GZIP.equals(oontentencoding) || CONTENT_ENCODING_XGZIP.equals(oontentencoding)) {
               lowLevelResponseContent =
                   new ConsumingInputStream(new GZIPInputStream(lowLevelResponseContent));
             }
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -26,6 +26,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.reflect.Type;
 import java.nio.charset.Charset;
+import java.util.Locale;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.zip.GZIPInputStream;
@@ -333,12 +334,12 @@ public final class HttpResponse {
         boolean contentProcessed = false;
         try {
           // gzip encoding (wrap content with GZipInputStream)
-          if (!returnRawInputStream
-              && this.contentEncoding != null
-              && (CONTENT_ENCODING_GZIP.equals(this.contentEncoding.trim())
-              || CONTENT_ENCODING_XGZIP.equals(this.contentEncoding.trim()))) {
-            lowLevelResponseContent =
-                new ConsumingInputStream(new GZIPInputStream(lowLevelResponseContent));
+          if (!returnRawInputStream && this.contentEncoding != null) {
+            final String oontentEncoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
+            if (CONTENT_ENCODING_GZIP.equals(oontentEncoding) || CONTENT_ENCODING_XGZIP.equals(oontentEncoding)) {
+              lowLevelResponseContent =
+                  new ConsumingInputStream(new GZIPInputStream(lowLevelResponseContent));
+            }
           }
           // logging (wrap content with LoggingInputStream)
           Logger logger = HttpTransport.LOGGER;
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -83,7 +83,7 @@ public final class HttpResponse {
   /** Content encoding for GZip */
   private static final String CONTENT_ENCODING_GZIP = "gzip";
 
-  /** Content encoding for GZip HTTP 1.1 */
+  /** Content encoding for GZip (legacy) */
   private static final String CONTENT_ENCODING_XGZIP = "x-gzip";
 
   /**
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -80,6 +80,12 @@
   /** Whether {@link #getContent()} should return raw input stream. */
   private final boolean returnRawInputStream;
 
+  /** Content encoding for GZip */
+  private static final String CONTENT_ENCODING_GZIP = "gzip";
+
+  /** Content encoding for GZip HTTP 1.1 */
+  private static final String CONTENT_ENCODING_XGZIP = "x-gzip";
```
elharo: Neat. Didn't know about this one. For the comment though, this is a legacy value, not the recommended value in HTTP 1.1.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -330,7 +336,8 @@ public InputStream getContent() throws IOException {
           String contentEncoding = this.contentEncoding;
           if (!returnRawInputStream
               && contentEncoding != null
-              && contentEncoding.contains("gzip")) {
+              && (CONTENT_ENCODING_GZIP.equalsIgnoreCase(contentEncoding.trim())
```
elharo: equalsIgnoreCase needs a character set; in this context UTF-8 probably works.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -330,7 +336,8 @@ public InputStream getContent() throws IOException {
           String contentEncoding = this.contentEncoding;
```
elharo: Existing i\ssue note: I don't see why we copy this into a local variable.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
egsavage: Please see latest commit.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
egsavage: I'm not sure where your concerns lie since Java Strings are UTF-16. Please advise.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
egsavage: Please see the latest commit.

egsavage:
> This needs a test case.

Please see the latest commit.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -333,12 +333,12 @@ public InputStream getContent() throws IOException {
...
         try {
           // gzip encoding (wrap content with GZipInputStream)
-          String contentEncoding = this.contentEncoding;
           if (!returnRawInputStream
-              && contentEncoding != null
-              && (CONTENT_ENCODING_GZIP.equalsIgnoreCase(contentEncoding.trim())
-                  || CONTENT_ENCODING_XGZIP.equalsIgnoreCase(contentEncoding.trim()))) {
-            lowLevelResponseContent = new GZIPInputStream(lowLevelResponseContent);
+                  && this.contentEncoding != null
+                  && (CONTENT_ENCODING_GZIP.equalsIgnoreCase(this.contentEncoding.trim())
```
elharo: needs a locale; case transformation depends on locale

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
```diff
@@ -495,4 +495,30 @@ public LowLevelHttpResponse execute() throws IOException {
...
   }
+
+  public void testGetContent_otherEncodingWithgzipInItsName_GzipIsNotUsed() throws IOException {
+    final MockLowLevelHttpResponse mockResponse = new MockLowLevelHttpResponse();
+    mockResponse.setContent("abcd");
+    mockResponse.setContentEncoding("otherEncodingWithgzipInItsName");
+    mockResponse.setContentType("text/plain");
+
+    HttpTransport transport =
+            new MockHttpTransport() {
```
elharo: google style is two space indents with 4 spaces for a continuation line

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -334,11 +334,11 @@ public InputStream getContent() throws IOException {
         try {
           // gzip encoding (wrap content with GZipInputStream)
           if (!returnRawInputStream
-                  && this.contentEncoding != null
-                  && (CONTENT_ENCODING_GZIP.equalsIgnoreCase(this.contentEncoding.trim())
-                      || CONTENT_ENCODING_XGZIP.equalsIgnoreCase(this.contentEncoding.trim()))) {
+              && this.contentEncoding != null
+              && (CONTENT_ENCODING_GZIP.equals(this.contentEncoding.trim())
```
elharo: I think per spec you need ignore case--correct me if I'm wrong about that--but you need to do it with the right locale, not necessarily the default.

chingor13: The header name is case-insensitive, header values are dependent on which field. For content encoding, valid values are a list of defined tokens [maintained by IANA](http://www.iana.org/assignments/http-parameters/http-parameters.xhtml#content-coding).

egsavage: @elharo , @chingor13 - I have another commit that has case-insensitive checks and hopefully addresses the prior locale issues mentioned. But given the recent point raised by @chingor13 , it seems that only the header NAME should be case-insensitive and not the actual values which I believe the current PR addresses.

Let me know if you'd like me to push the new commit to the branch.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -333,12 +334,12 @@ public InputStream getContent() throws IOException {
...
         try {
           // gzip encoding (wrap content with GZipInputStream)
-          if (!returnRawInputStream
-              && this.contentEncoding != null
-              && (CONTENT_ENCODING_GZIP.equals(this.contentEncoding.trim())
-              || CONTENT_ENCODING_XGZIP.equals(this.contentEncoding.trim()))) {
-            lowLevelResponseContent =
-                new ConsumingInputStream(new GZIPInputStream(lowLevelResponseContent));
+          if (!returnRawInputStream && this.contentEncoding != null) {
+            final String oontentEncoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
```
elharo: google style doesn't use final on local variables except where required;

also oontent --> content

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
```diff
@@ -461,6 +462,21 @@ public LowLevelHttpResponse execute() throws IOException {
   }
 
   public void testGetContent_gzipEncoding_finishReading() throws IOException {
+    do_testGetContent_gzipEncoding_finishReading("gzip");
+  }
+
+  public void testGetContent_gzipEncoding_finishReadingWithUppercaseContentEncoding() throws IOException {
```
elharo: needs `@Test`; writing the test first and making sure it fails helps avoid problems like this

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
```diff
@@ -461,6 +462,21 @@ public LowLevelHttpResponse execute() throws IOException {
...
+    do_testGetContent_gzipEncoding_finishReading("GZIP");
+  }
+
+  public void testGetContent_gzipEncoding_finishReadingWithDifferentDefaultLocaleAndUppercaseContentEncoding() throws IOException {
```
elharo: needs `@Test`;

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
```diff
@@ -461,6 +462,21 @@ public LowLevelHttpResponse execute() throws IOException {
...
+    Locale originalDefaultLocale = Locale.getDefault();
+    Locale.setDefault(Locale.JAPAN);
+    do_testGetContent_gzipEncoding_finishReading("GZIP");
+    Locale.setDefault(originalDefaultLocale);
+  }
+
+  private void do_testGetContent_gzipEncoding_finishReading(String contentEncoding) throws IOException {
     byte[] dataToCompress = "abcd".getBytes(StandardCharsets.UTF_8);
     byte[] mockBytes;
```
elharo: you can probably push this (and more) into the try block

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
egsavage: This is an existing old test class using JUnit 3 method conventions thus the lack of the @Test annotation. It runs fine as-is.

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
```diff
@@ -461,6 +462,21 @@ public LowLevelHttpResponse execute() throws IOException {
...
+
+  public void testGetContent_gzipEncoding_finishReadingWithUppercaseContentEncoding() throws IOException {
+    do_testGetContent_gzipEncoding_finishReading("GZIP");
+  }
+
+  public void testGetContent_gzipEncoding_finishReadingWithDifferentDefaultLocaleAndUppercaseContentEncoding() throws IOException {
+    Locale originalDefaultLocale = Locale.getDefault();
+    Locale.setDefault(Locale.JAPAN);
+    do_testGetContent_gzipEncoding_finishReading("GZIP");
+    Locale.setDefault(originalDefaultLocale);
```
elharo: this probably should be in a finally block so it's still reset if the test fails

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
```diff
@@ -461,6 +462,21 @@ public LowLevelHttpResponse execute() throws IOException {
...
+    do_testGetContent_gzipEncoding_finishReading("gzip");
+  }
+
+  public void testGetContent_gzipEncoding_finishReadingWithUppercaseContentEncoding() throws IOException {
+    do_testGetContent_gzipEncoding_finishReading("GZIP");
+  }
+
+  public void testGetContent_gzipEncoding_finishReadingWithDifferentDefaultLocaleAndUppercaseContentEncoding() throws IOException {
+    Locale originalDefaultLocale = Locale.getDefault();
+    Locale.setDefault(Locale.JAPAN);
```
elharo: Try Turkish instead:
```
Locale turkish = Locale.forLanguageTag("tr-TR");
```

That's the one common language where i and I do not case like they do in English.

egsavage: I switched it to Turkish, but JAPAN in the existing code would have given an exception without the locale specification in the implementation.

PR #846, opened 2019-10-15
fix: make opencensus an optional dependency

codyoss:
With these changes the user has the option to use opencensus. If
the deps are not on the classpath noop impls will be used. In the
future opencensus will be ripped out in favor of opentelemetry.

As a part of these changes OpenCensusUtils was moved to a
package-private scope.

Fixes: #621

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/OpenCensusScope.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/OpenCensusScope.java
@@ -1,24 +1,19 @@
 /**
  * Copyright 2019 Google LLC
  *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
+ * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * except in compliance with the License. You may obtain a copy of the License at
  *
- *     https://www.apache.org/licenses/LICENSE-2.0
+ * <p>https://www.apache.org/licenses/LICENSE-2.0
  *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
+ * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.google.api.client.http;
 
-/**
- * A thin wrapper around an OpenCensus Scope with some convenience methods.
- */
+/** A thin wrapper around an OpenCensus Scope with some convenience methods. */
 final class OpenCensusScope implements Scope {
   private final io.opencensus.common.Scope scope;
 
--- a/google-http-client/src/main/java/com/google/api/client/http/OpenCensusScope.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/OpenCensusScope.java
@@ -1,12 +1,12 @@
 /**
  * Copyright 2019 Google LLC
  *
- * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
  * except in compliance with the License. You may obtain a copy of the License at
  *
- * <p>https://www.apache.org/licenses/LICENSE-2.0
+ * https://www.apache.org/licenses/LICENSE-2.0
  *
- * <p>Unless required by applicable law or agreed to in writing, software distributed under the
+ * Unless required by applicable law or agreed to in writing, software distributed under the
  * License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
  * express or implied. See the License for the specific language governing permissions and
  * limitations under the License.
--- a/google-http-client/src/main/java/com/google/api/client/http/OpenCensusScope.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/OpenCensusScope.java
@@ -1,0 +1,33 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


codyoss:
@chingor13 I think this works. Before putting more work into this though I wanted to get feedback if we do want to do this. I wonder if things that depend on this lib expect opencensus to be there creating span and the likes... I am not sure if there is a good way for us to tell this.

codyoss:
hmmm, this still does need some more work. I will try to get it working today...

google-http-client/src/main/java/com/google/api/client/http/OpenCensusSpan.java
```diff
@@ -0,0 +1,83 @@
...
+
+  OpenCensusSpan() {
+    this.span = OpenCensusUtils.getTracer()
+        .spanBuilder(OpenCensusUtils.SPAN_NAME_HTTP_REQUEST_EXECUTE)
+        .setRecordEvents(OpenCensusUtils.isRecordEvent())
+        .startSpan();
+  }
+
+  @Override
+  public final void addCommonHTTPAttributes(String requestMethod, String host, String path, String url) {
```
elharo: addCommonHttpAttributes per google naming conventions

google-http-client/src/main/java/com/google/api/client/http/ScopeFactory.java
```diff
@@ -0,0 +1,54 @@
...
+
+  static Scope getScope(Span span) {
+    if (HAS_OPEN_CENSUS_CLASSES_ON_CLASSPATH) {
+      return new OpenCensusScope((OpenCensusSpan) span);
+    } else {
+      return noopScope;
+    }
+  }
+
+  private static boolean hasClasses(String ...classNames) {
```
elharo: not sure; is this usually String... classNames?

google-http-client/src/main/java/com/google/api/client/http/ScopeFactory.java
```diff
@@ -0,0 +1,54 @@
...
+    for (String className : classNames) {
+      try {
+        Class.forName(className, false, ScopeFactory.class.getClassLoader());
+      } catch (Throwable t) {
```
elharo: This should be more specific

google-http-client/src/main/java/com/google/api/client/http/ScopeFactory.java
codyoss: Would you prefer `List`?

elharo: The question was whether the space goes before or after the three dots by convention

codyoss: Awe I see... you are right. I ran formatted on all the new files. I filed an issue in this repo yesterday to get that to be a part of the build. A lot of files need to be formatted. Good catch!

google-http-client/src/main/java/com/google/api/client/http/OpenCensusUtils.java
```diff
@@ -34,14 +33,12 @@
...
 /**
- * {@link Beta} <br>
  * Utilities for Census monitoring and tracing.
  *
  * @author Hailong Wen
  * @since 1.28
  */
-@Beta
-public class OpenCensusUtils {
+class OpenCensusUtils {
```
chingor13: This looks like a breaking change, even though it's annotated `@Beta`. Is this intentional?

google-http-client/src/main/java/com/google/api/client/http/ScopeFactory.java
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+package com.google.api.client.http;
+
+class ScopeFactory {
+  /** This is used to check if the optional OpenCensus deps are on the classpath */
+  private static final boolean HAS_OPEN_CENSUS_CLASSES_ON_CLASSPATH =
+      hasClasses(
+          "io.opencensus.trace.Span",
+          "io.opencensus.implcore.trace.RecordEventsSpanImpl",
```
chingor13: Do we actually need to check for this class? `opencensus-api` already ships with a no-op implementation by default, unless `opencensus-impl` is loaded.

google-http-client/src/main/java/com/google/api/client/http/OpenCensusSpan.java
```diff
@@ -0,0 +1,80 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
```
elharo: You don't need `<p>` tags in the license

google-http-client/src/main/java/com/google/api/client/http/OpenCensusScope.java
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
```
elharo: You don't need `<p>` tags in the license

google-http-client/src/main/java/com/google/api/client/http/Scope.java
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright 2019 Google LLC
+ *
+ * <p>Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file
```
elharo: You don't need `<p>` tags in the license

google-http-client/src/main/java/com/google/api/client/http/OpenCensusUtils.java
codyoss: Yes it was. This class imports opencensus so its usage I think needs to be guarded by the class loading check. I figured since it was beta it was okay to move it out of the public scope and make it package private. Is that not the case? If we don't shift this to an internal class I don't think we can make the dependencies optional, right?

google-http-client/src/main/java/com/google/api/client/http/Scope.java
codyoss: Whoops, it appears the formatter got overzealous. I will fill all of these.

google-http-client/src/main/java/com/google/api/client/http/ScopeFactory.java
codyoss: I believe you are right, I will remove this one from both factories

google-http-client/src/main/java/com/google/api/client/http/OpenCensusUtils.java
codyoss: @chingor13 WDYT?

elharo: "Features marked with the @Beta annotation at the class or method level are subject to change. They might be modified in any way, or even removed, in any major release. You should not use beta features if your code is a library itself (that is, if your code is used on the CLASSPATH of users outside your own control)."

codyoss:
I think after talking to @chingor13 offline, we decided to not move forward with this. Closing. We will wait take the version bump when this dependency is replaced/removed.

PR #851, opened 2019-10-18
remove fields deprecated 4 years ago

elharo:
@chingor13


elharo:
Ping. Do we want to do this? Does this require a major version bump or not?

PR #853, opened 2019-10-21
depend on grpc-context

elharo:
This dependency is needed to keep OpenCensus from pulling in gRPC 1.22.1.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -241,7 +241,7 @@
         <artifactId>j2objc-annotations</artifactId>
         <version>1.3</version>
       </dependency>
-      <!-- without this OpenCensus uses an older version of grpc-context -->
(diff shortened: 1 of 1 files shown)


codyoss:
I know we have not merged #846 yet, but if we do, I think think this should be made optional as well.

chingor13:
I don't think this belongs here - there is nothing related to gRPC in this repo. A user of this library who uses gRPC should be configuring their grpc versions.

elharo:
The problem is OpenCensus is not up to date and is not going to be. A user who does not use OpenCensus or gRPC directly can still hit dependency conflicts due to this. The alternative is ripping OpenCensus out again until OpenTelemetry is ready.

lesv:
I agree w/ @elharo - we should talk about this on Wednesday.

PR #865, opened 2019-10-31
Fix various non-functional checkstyle warnings

Anusien:
Fixes the following checkstyle warnings:

> [WARN] /home/anusien/opensource/google-http-java-client/google-http-client/src/main/java/com/google/api/client/util/ArrayValueMap.java:14: 'package' should be separated from previous statement. [EmptyLineSeparator]

> [WARN] /home/anusien/opensource/google-http-java-client/google-http-client/src/main/java/com/google/api/client/http/GZipEncoding.java:45: Empty catch block. [EmptyCatchBlock]

> [WARN] /home/anusien/opensource/google-http-java-client/google-http-client/src/main/java/com/google/api/client/testing/util/TestableByteArrayInputStream.java:44:47: Array brackets at illegal position. [ArrayTypeStyle]

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/GZipEncoding.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/GZipEncoding.java
@@ -43,6 +43,7 @@ public class GZipEncoding implements HttpEncoding {
             try {
               flush();
             } catch (IOException ignored) {
+                // Nothing else we can do here
             }
           }
         };
--- a/google-http-client/src/main/java/com/google/api/client/testing/util/TestableByteArrayInputStream.java
+++ b/google-http-client/src/main/java/com/google/api/client/testing/util/TestableByteArrayInputStream.java
@@ -41,7 +41,7 @@ public class TestableByteArrayInputStream extends ByteArrayInputStream {
    * @param offset offset in the buffer of the first byte to read
    * @param length maximum number of bytes to read from the buffer
    */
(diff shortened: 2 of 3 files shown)


google-http-client/src/main/java/com/google/api/client/http/GZipEncoding.java
```diff
@@ -43,6 +43,7 @@ public void close() throws IOException {
             try {
               flush();
             } catch (IOException ignored) {
+                // Nothing else we can do here
```
elharo: I'm not sure this is true. I wonder why we don't propagate the IOException in this case?

Anusien:
@googlebot I signed it!

google-http-client/src/main/java/com/google/api/client/http/GZipEncoding.java
Anusien: I have no clue.

chanseokoh:
```suggestion
                // fall through
```

google-http-client/src/main/java/com/google/api/client/http/GZipEncoding.java
```diff
@@ -43,6 +43,7 @@ public void close() throws IOException {
             try {
               flush();
             } catch (IOException ignored) {
+                // fall through
```
chanseokoh, line 46:
```suggestion
              // fall through
```

PR #876, opened 2019-11-04
fix: redirect on 308 (Permanent Redirect) too

### LINKED ISSUE #873: setFollowRedirects(true) does not follow 308 (Permanent Redirect)
Using the latest versions:
google-http-client: 1.33.0
google-http-client-apache-v2: 1.33.0

By default `setFollowRedirects(true)` is set, and the library automatically follows the following redirects.

301 (Moved Permanently)
302 (Found)
303 (See Other)
307 (Temporary Redirect)

However, 308 (Permanent Redirect) isn't followed from my [experiment](https://github.com/googleapis/google-http-java-client/pull/871#issuecomment-548913894).

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#3xx_Redirection says

> 307 and 308 parallel the behaviors of 302 and 301

so I reckon if the library is following 301, 302, and 307, perhaps it should follow 308 too.

chanseokoh:
Fixes #873.

### DIFF
--- a/google-http-client/src/test/java/com/google/api/client/http/HttpStatusCodesTest.java
+++ b/google-http-client/src/test/java/com/google/api/client/http/HttpStatusCodesTest.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.api.client.http;
+
+import junit.framework.TestCase;
+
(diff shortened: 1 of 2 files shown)


google-http-client/src/test/java/com/google/api/client/http/HttpStatusCodesTest.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2019 Google Inc.
```
elharo: no (c)

also LLC

google-http-client/src/test/java/com/google/api/client/http/HttpStatusCodesTest.java
```diff
@@ -0,0 +1,35 @@
...
+package com.google.api.client.http;
+
+import junit.framework.TestCase;
+
+/** Tests {@link HttpStatusCodes}. */
+public class HttpStatusCodesTest extends TestCase {
+
+  public void testIsRedirect_3xx() {
+    assertTrue(HttpStatusCodes.isRedirect(301));
+    assertTrue(HttpStatusCodes.isRedirect(302));
```
elharo: what about 309 to 399? These are all undefined redirect codes according to HTTP.

chanseokoh: I think the method name should really be `isDefinedRedirect` or such, given where this code is used to handle automatic redirection. The Javadoc at least describes the behavior correctly.

PR #895, opened 2019-11-25
fix millisecond parsing errors

### LINKED ISSUE #894: One second off with pre epoch dates
second test below fails. This was probably introduced in #752

```
public void testEpoch() {
    assertParsedRfc3339(
        "1970-01-01T00:00:00.000Z", SecondsAndNanos.ofSecondsAndNanos(0, 0));
  }

  public void testOneSecondBeforeEpoch() {
    assertParsedRfc3339(
        "1969-12-31T23:59:59.000Z", SecondsAndNanos.ofSecondsAndNanos(-1, 0));
  }
```

elharo:
Fixes #894 @suztomo @olavloite @codyoss

### DIFF
--- a/google-http-client/src/test/java/com/google/api/client/util/DateTimeTest.java
+++ b/google-http-client/src/test/java/com/google/api/client/util/DateTimeTest.java
@@ -15,6 +15,7 @@
 package com.google.api.client.util;
 
 import com.google.api.client.util.DateTime.SecondsAndNanos;
+
 import java.util.Date;
 import java.util.TimeZone;
 import junit.framework.TestCase;
@@ -27,12 +28,6 @@ import junit.framework.TestCase;
 public class DateTimeTest extends TestCase {
 
   private TimeZone originalTimeZone;
-
-  public DateTimeTest() {}
-
-  public DateTimeTest(String testName) {
-    super(testName);
-  }
 
   @Override
   protected void setUp() throws Exception {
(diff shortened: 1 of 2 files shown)


google-http-client/src/test/java/com/google/api/client/util/DateTimeTest.java
```diff
@@ -225,8 +220,18 @@ public void testParseRfc3339ToSecondsAndNanos() {
...
+  
+  public void testEpoch() {
+    assertParsedRfc3339(
+        "1970-01-01T00:00:00.000Z", SecondsAndNanos.ofSecondsAndNanos(0, 0));
+  }
+
+  public void testOneSecondBeforeEpoch() {
+    assertParsedRfc3339(
+        "1969-12-31T23:59:59.000Z", SecondsAndNanos.ofSecondsAndNanos(-1, 0));
+  }
```
suztomo: Would you add the following test too?

```
public void testGregorianCalendarStartDate() {
    assertParsedRfc3339(
        "1582-10-15T01:23:45.123Z", SecondsAndNanos.ofSecondsAndNanos(-12219287774L, -877000000));
}
```
(not 100% sure about the value of nanosecond part to represent `-12219287774877` milliseconds)

As per Beam's PubsubClientTest, date 1582-10-15 is an important date when Gregorian Calendar was adopted.
https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/test/java/org/apache/beam/sdk/io/gcp/pubsub/PubsubClientTest.java#L144

google-http-client/src/test/java/com/google/api/client/util/DateTimeTest.java
```diff
@@ -15,6 +15,7 @@
 package com.google.api.client.util;
 
 import com.google.api.client.util.DateTime.SecondsAndNanos;
+
```
suztomo: Nit?

elharo: done

google-http-client/src/test/java/com/google/api/client/util/DateTimeTest.java
elharo: I can't verify that this value is correct, so I don't want to add it here.

suztomo: I will try to add this in a separate PR then.

PR #910, opened 2019-12-10
feat: support chunked transfer encoding

### LINKED ISSUE #648: HttpContent is encoded two times when posting data

codyoss:
Currently any time HttpRequest works with encoded data it encodes
the data twice. Once for the actual stream and once for checking
the length of the stream. Instead, when there is encoding just don't
set the content length. This will cause the underlying transports,
with a few tweaks, to use Transfer-Encoding: chunked.

Fixes #648

### DIFF
--- a/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpRequest.java
+++ b/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpRequest.java
@@ -23,9 +23,7 @@ import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
 import org.apache.http.client.methods.HttpRequestBase;
 
-/**
- * @author Yaniv Inbar
- */
+/** @author Yaniv Inbar */
 final class ApacheHttpRequest extends LowLevelHttpRequest {
   private final HttpClient httpClient;
 
@@ -38,11 +36,12 @@ final class ApacheHttpRequest extends LowLevelHttpRequest {
     this.httpClient = httpClient;
     this.request = request;
     // disable redirects as google-http-client handles redirects
-    this.requestConfig = RequestConfig.custom()
-        .setRedirectsEnabled(false)
-        .setNormalizeUri(false)
-        // TODO(chingor): configure in HttpClientBuilder when available
-        .setStaleConnectionCheckEnabled(false);
+    this.requestConfig =
+        RequestConfig.custom()
+            .setRedirectsEnabled(false)
+            .setNormalizeUri(false)
+            // TODO(chingor): configure in HttpClientBuilder when available
+            .setStaleConnectionCheckEnabled(false);
   }
 
   @Override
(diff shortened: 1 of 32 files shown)


codyoss:
Sorry for some whitespace churn. I ran the formatter for my commit, and it looks there were a handful of other files that were not complying.

google-http-client/src/main/java/com/google/api/client/testing/http/javanet/MockHttpURLConnection.java
```diff
@@ -201,4 +201,8 @@ public String getHeaderField(String name) {
     List<String> values = headers.get(name);
     return values == null ? null : values.get(0);
   }
+
+  public int getChunckLength() {
```
chingor13:
```suggestion
  public int getChunkLength() {
```

codyoss:
Sure thing.

google-http-client-apache-v2/src/test/java/com/google/api/client/http/apache/v2/ApacheHttpRequestTest.java
```diff
@@ -0,0 +1,45 @@
+package com.google.api.client.http.apache.v2;
```
elharo: copyright

codyoss:
That should be better, only formatted the files I changed. Sorry about that.

google-http-client/src/main/java/com/google/api/client/http/apache/ApacheHttpRequest.java
```diff
@@ -54,12 +52,16 @@ public void setTimeout(int connectTimeout, int readTimeout) throws IOException {
   @Override
   public LowLevelHttpResponse execute() throws IOException {
     if (getStreamingContent() != null) {
-      Preconditions.checkArgument(request instanceof HttpEntityEnclosingRequest,
+      Preconditions.checkArgument(
+          request instanceof HttpEntityEnclosingRequest,
```
elharo: request is not an argument here. This should be checkState

google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpRequest.java
```diff
@@ -52,19 +51,22 @@ public void addHeader(String name, String value) {
...
-    requestConfig.setConnectTimeout(connectTimeout)
-        .setSocketTimeout(readTimeout);
+    requestConfig.setConnectTimeout(connectTimeout).setSocketTimeout(readTimeout);
   }
 
   @Override
   public LowLevelHttpResponse execute() throws IOException {
     if (getStreamingContent() != null) {
-      Preconditions.checkArgument(request instanceof HttpEntityEnclosingRequest,
+      Preconditions.checkArgument(
```
elharo: checkState

google-http-client/src/test/java/com/google/api/client/http/javanet/NetHttpRequestTest.java
```diff
@@ -203,4 +203,34 @@ public void close() throws IOException {
...
+    MockHttpURLConnection connection = new MockHttpURLConnection(new URL(HttpTesting.SIMPLE_URL));
+    connection.setRequestMethod("POST");
+    NetHttpRequest request = new NetHttpRequest(connection);
+    InputStream is = NetHttpRequestTest.class.getClassLoader().getResourceAsStream("file.txt");
+    HttpContent content = new InputStreamContent("text/plain", is);
+    request.setStreamingContent(content);
+    request.setContentEncoding("gzip");
+    request.execute();
+
+    assertEquals(connection.getChunkLength(), 4096);
```
elharo: expected value comes first so assertEquals(4096, connection.getChunkLength());

Issue #905, opened 2019-12-05
InvalidArgumentException from HttpResponse if the content-type is invalid


jmorrise:
I'm hitting an edge case at [HttpResponse.java#153](https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java#L153): if the contentType of the response is invalid, an InvalidArgumentException is thrown while constructing the media type. My code makes an http request to read the contents of a file from Google Cloud Storage, so if the owner of the file has set an invalid content-type (which is easy to do in GCS, unfortunately), I can't read the file at all.

I understand that this might just be WAI, and I could just throw the error back to the user, however it would be nice if I could handle this gracefully and read the file contents as if the mediaType were unknown. It's difficult for me to catch and handle the error myself since I'm not calling this library directly (another library that I don't own is making the http request).

Since mediaType is nullable anyway, would it be reasonable behavior to set it to null if the contentType were invalid? Something like

```java
mediaType = contentType == null || !HttpMediaType.isValid(contentType) ? null : new HttpMediaType(contentType);
```
where isValid would return true iff the contentType can be parsed to a valid mediaType.

Would this be an okay change? I wanted to check before actually sending a pull request. Thanks.

chingor13:
I think allowing this to be nullable is ok. It might be worth considering lazy loading the `mediaType` value as it's not critical to the contents of the instance. I'm not sure of a time when you'd want to get a `IllegalArgumentException` (a runtime, unchecked exception) when parsing an HTTP response without a way to inspect the response.

PR #911, opened 2019-12-10
fix: set mediaType to null if contentType cannot be parsed

jmorrise:
Fixes issue #905 by setting mediaType to null if the contentType is invalid.

### DIFF
--- a/google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
+++ b/google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
@@ -58,6 +58,8 @@ public class HttpResponseTest extends TestCase {
 
   private static final String SAMPLE = "123\u05D9\u05e0\u05D9\u05D1";
   private static final String SAMPLE2 = "123abc";
+  private static final String VALID_CONTENT_TYPE = "text/plain";
+  private static final String INVALID_CONTENT_TYPE = "!!!invalid!!!";
 
   public void testParseAsString_utf8() throws Exception {
     HttpTransport transport =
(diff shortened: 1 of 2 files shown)


google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
```diff
@@ -58,6 +58,8 @@ public void testParseAsString_none() throws Exception {
 
   private static final String SAMPLE = "123\u05D9\u05e0\u05D9\u05D1";
   private static final String SAMPLE2 = "123abc";
+  private static final String VALID_CONTENT_TYPE = "text/plain";
```
BenWhitehead: Can we add another valid complex content type? Maybe something like `application/vnd.com.google.datastore.entity+json; charset=utf-8; version=v1; q=0.9`

jmorrise: Okay, added one with parameters.

jmorrise:
I force-pushed to resolve the conventionalcommits error, but I'm afraid things might be in a weird state now - let me know if I need to fix this.

google-http-client/src/test/java/com/google/api/client/http/HttpResponseTest.java
BenWhitehead: Awesome, thanks!

chingor13:
Note that this is potentially a behavior breaking change in that previously you could get an `InvalidArgumentException` (unchecked, runtime exception) when executing a request. Now, you will get the response with a null (unparsed) `MediaType` which would seem to be always preferred anyways.

PR #912, opened 2019-12-10
feat: decode uri path components correctly

### LINKED ISSUE #398: GenericUrl converts plus sign into space within URI path component

codyoss:
The old implementation was incorrectly treating '+' as a space. Now
the only things that get decoded in the path are uri escaped sequences.

Fixes #398

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java
@@ -16,6 +16,8 @@ package com.google.api.client.util.escape;
 
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
+import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
 
 /**
  * Utility functions for dealing with {@code CharEscaper}s, and some commonly used {@code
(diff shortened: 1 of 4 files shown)


google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java
```diff
@@ -90,6 +92,59 @@ public static String decodeUri(String uri) {
...
+   */
+  public static String decodeUriPath(String path) {
+    if (path == null) {
+      return null;
+    }
+    ByteBuffer buf = null;
+    int length = path.length();
+    StringBuilder sb = new StringBuilder(length);
+
+    char c;
```
elharo: this should be inside the for loop

google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java
```diff
@@ -90,6 +92,59 @@ public static String decodeUri(String uri) {
...
+    for (int i = 0; i < length; i++) {
+      c = path.charAt(i);
+      if (c == '%') {
+        if (i + 2 < length) {
+          if (buf == null) {
+            buf = ByteBuffer.allocate(Integer.SIZE / 8);
```
elharo: you could initialize this in line 109 and avoid the if-else here

PR #913, opened 2019-12-10
feat: decode uri path components correctly

### LINKED ISSUE #398: GenericUrl converts plus sign into space within URI path component

codyoss:
The old implementation was incorrectly treating '+' as a space. Now
the only things that get decoded in the path are uri escaped sequences.

Fixes #398

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/GenericUrl.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/GenericUrl.java
@@ -81,10 +81,10 @@ public class GenericUrl extends GenericData {
   private String fragment;
 
   /**
-    * If true, the URL string originally given is used as is (without encoding, decoding and
-    * escaping) whenever referenced; otherwise, part of the URL string may be encoded or decoded as
-    * deemed appropriate or necessary.
-    */
+   * If true, the URL string originally given is used as is (without encoding, decoding and
+   * escaping) whenever referenced; otherwise, part of the URL string may be encoded or decoded as
+   * deemed appropriate or necessary.
+   */
   private boolean verbatim;
 
   public GenericUrl() {}
@@ -112,19 +112,19 @@ public class GenericUrl extends GenericData {
   /**
    * Constructs from an encoded URL.
    *
-   * <p>Any known query parameters with pre-defined fields as data keys are parsed based on
-   * their data type. Any unrecognized query parameter are always parsed as a string.
+   * <p>Any known query parameters with pre-defined fields as data keys are parsed based on their
+   * data type. Any unrecognized query parameter are always parsed as a string.
    *
    * <p>Any {@link MalformedURLException} is wrapped in an {@link IllegalArgumentException}.
    *
    * @param encodedUrl encoded URL, including any existing query parameters that should be parsed
-   * @param verbatim flag, to specify if URL should be used as is (without encoding, decoding and escaping)
+   * @param verbatim flag, to specify if URL should be used as is (without encoding, decoding and
+   *     escaping)
    * @throws IllegalArgumentException if URL has a syntax error
    */
   public GenericUrl(String encodedUrl, boolean verbatim) {
     this(parseURL(encodedUrl), verbatim);
   }
-
 
   /**
    * Constructs from a URI.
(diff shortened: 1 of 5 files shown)


google-http-client/src/main/java/com/google/api/client/http/GenericUrl.java
```diff
@@ -608,13 +611,17 @@ private void appendRawPathFromParts(StringBuilder buf) {
...
-  static void addQueryParams(Set<Entry<String, Object>> entrySet, StringBuilder buf, boolean verbatim) {
+  static void addQueryParams(
+      Set<Entry<String, Object>> entrySet, StringBuilder buf, boolean verbatim) {
     // (similar to UrlEncodedContent)
     boolean first = true;
     for (Map.Entry<String, Object> nameValueEntry : entrySet) {
       Object value = nameValueEntry.getValue();
       if (value != null) {
-        String name =  verbatim ? nameValueEntry.getKey() : CharEscapers.escapeUriQuery(nameValueEntry.getKey());
+        String name =
```
elharo: I preferred the original one line version here.

google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java
```diff
@@ -90,6 +92,28 @@ public static String decodeUri(String uri) {
...
+   *
+   * @param path the value to be decoded
+   * @return decoded version of {@code path}
+   */
+  public static String decodeUriPath(String path) {
+    if (path == null) {
+      return null;
+    }
+    try {
+      return URLDecoder.decode(path.replace("+", "%2B"), "UTF-8");
```
elharo: I think you can use Charsets.UTF_8 or some such and avoid the try catch

google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java
```diff
@@ -90,6 +92,28 @@ public static String decodeUri(String uri) {
     }
   }
 
+  /**
+   * Decodes the path component of a URI. This must be done via a method that does not try to
+   * convert + into spaces(the behavior of {@link java.net.URLDecoder#decode(String, String)}). This
+   * method will transform URI encoded value into their decoded symbols.
```
elharo: will transform --> transforms URI encoded values

google-http-client/src/main/java/com/google/api/client/http/GenericUrl.java
codyoss: I don't disagree, but this with what running `mvn com.coveo:fmt-maven-plugin:format` yields .

google-http-client/src/main/java/com/google/api/client/util/escape/CharEscapers.java
codyoss: Charset api requires 10+.

elharo: StandardCharsets.UTF_8 is in java 7.

codyoss: I meant the api that accepts a Charset, but I suppose I can call `name()` on it

dmitry-fa:
This change caused the problem with using space in object names in java storage:

https://github.com/googleapis/java-storage/issues/53
https://github.com/googleapis/java-storage/issues/57

athakor:
@dmitry-fa correct, one more issue related to this changes.
https://github.com/googleapis/java-storage/issues/121

mihalyr:
This is the suspected cause of Google Sheets V4 failures when the sheet title includes a plus sign: https://issuetracker.google.com/issues/217399871

mihalyr:
Hi, in https://issuetracker.google.com/issues/217399871 we have noticed a problem with Google Sheets API when using `google-http-client` version `1.34.0` or older that includes this change. The problem is handling `+` signs in sheet titles that lead to API failures when trying to append cells to a sheet that contains `+` in the title. This was working well before this change. 

It is a bit confusing as the error message we get shows that the problem is that the sheet title is passed in the URL and the plus gets replaced by a space, while the value range in the body remains unchanged and this mismatch results in API failures that prevent adding cells to such sheets.

PR #916, opened 2019-12-16
fix: use random UUID for multipart boundary delimiter

### LINKED ISSUE #909: MultipartContent class uses the same simple string as the default boundary
`com.google.api.client.http.MultipartContent` class by default uses constant '`__END_OF_PART__`'  string as the boundary (part delimiter).
It will cause problem if the data to be transfered includes this string. Such problems could be really hard to investigate.
Example of such problem:
https://github.com/googleapis/google-cloud-java/issues/6386

The  class also provides `setBoundary(String)` method to alter the default boundary, but this method is not accessible from the other project.

dmitry-fa:
Fixes #909 
Fixes https://github.com/googleapis/google-cloud-java/issues/6386

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/MultipartContent.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/MultipartContent.java
@@ -31,7 +31,7 @@ import java.util.Collections;
  * and <a href="http://tools.ietf.org/html/rfc1521#section-7.2.2">RFC 2046: Multipurpose Internet
  * Mail Extensions: The Multipart/mixed (primary) subtype</a>.
  *
- * <p>By default the media type is {@code "multipart/related; boundary=__END_OF_PART__"}, but this
+ * <p>By default the media type is {@code "multipart/related; boundary=__END_OF_PART__<random UUID>__"}, but this
  * may be customized by calling {@link #setMediaType(HttpMediaType)}, {@link #getMediaType()}, or
  * {@link #setBoundary(String)}.
  *
@@ -47,10 +47,14 @@ public class MultipartContent extends AbstractHttpContent {
   private static final String TWO_DASHES = "--";
 
   /** Parts of the HTTP multipart request. */
-  private ArrayList<Part> parts = new ArrayList<Part>();
+  private ArrayList<Part> parts = new ArrayList<>();
 
   public MultipartContent() {
-    super(new HttpMediaType("multipart/related").setParameter("boundary", "__END_OF_PART__"));
+    this("__END_OF_PART__" + java.util.UUID.randomUUID().toString() + "__");
+  }
+
+  public MultipartContent(String boundary) {
+    super(new HttpMediaType("multipart/related").setParameter("boundary", boundary));
   }
 
   public void writeTo(OutputStream out) throws IOException {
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/http/MultipartContent.java
```diff
@@ -47,10 +47,14 @@
   private static final String TWO_DASHES = "--";
 
   /** Parts of the HTTP multipart request. */
-  private ArrayList<Part> parts = new ArrayList<Part>();
+  private ArrayList<Part> parts = new ArrayList<>();
 
   public MultipartContent() {
-    super(new HttpMediaType("multipart/related").setParameter("boundary", "__END_OF_PART__"));
+    this("__END_OF_PART__" + java.util.UUID.randomUUID().toString() + "__");
```
elharo: Is there a reason not to import java.util.UUID? e.g a conflict with a similarly named class?

google-http-client/src/test/java/com/google/api/client/http/MultipartContentTest.java
```diff
@@ -26,55 +26,76 @@
...
+    assertTrue(boundaryString.startsWith(BOUNDARY));
+    assertTrue(boundaryString.endsWith("__"));
+    assertEquals("multipart/related; boundary=" + boundaryString, content.getType());
+
+    final String[][] VALUES = new String[][] {
+            {"Hello world", "text/plain"},
+            {"<xml>Hi</xml>", "application/xml"},
+            {"{x:1,y:2}", "application/json"}
+    };
+    StringBuilder expectedContentSB = new StringBuilder();
```
elharo: no abbreviations per google style, but simply "expectedContent" or even just expected is fine.

google-http-client/src/test/java/com/google/api/client/http/MultipartContentTest.java
```diff
@@ -26,55 +26,76 @@
...
+      expectedContentSB.append("--").append(boundaryString).append(CRLF)
+              .append(headers(contentType, contentValue)).append(CRLF)
+              .append(contentValue).append(CRLF);
+    }
+    expectedContentSB.append("--").append(boundaryString).append("--").append(CRLF);
+    // write to string
+    ByteArrayOutputStream out = new ByteArrayOutputStream();
+    content.writeTo(out);
+    String expectedContent = expectedContentSB.toString();
+    assertEquals(expectedContent, out.toString());
```
elharo: out.toString needs to specify a locale

google-http-client/src/main/java/com/google/api/client/http/MultipartContent.java
dmitry-fa: fixed.

google-http-client/src/test/java/com/google/api/client/http/MultipartContentTest.java
dmitry-fa: expectedContent is used later. renamed to expectedStringBuilder

google-http-client/src/test/java/com/google/api/client/http/MultipartContentTest.java
dmitry-fa: fixed in two places. Thanks for catching that.

PR #927, opened 2019-12-28
fix: incorrect check for Windows OS in FileDataStoreFactory

### LINKED ISSUE #953: Incorrect check for Windows (case sensitivity)
The value for StandardSystemProperty.OS_NAME on win10 is "Windows 10" which fails the case sensitive startsWith check

See https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java#L55

diesieben07:
The "Windows" part of `os.name` is not always uppercase. On Windows 10 for example I get `"Windows 10"`, which does not pass the current check. This leads to `setPermissionsToOwnerOnly` instead of `setPermissionsToOwnerOnlyWindows` being executed, which then issues the following warning (and does nothing): `Unable to set permissions for C:\example, because you are running on a non-POSIX file system`

Fixes #953

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java
@@ -53,7 +53,7 @@ public class FileDataStoreFactory extends AbstractDataStoreFactory {
   private static final Logger LOGGER = Logger.getLogger(FileDataStoreFactory.class.getName());
 
   private static final boolean IS_WINDOWS = StandardSystemProperty.OS_NAME.value()
-      .startsWith("WINDOWS");
+      .regionMatches(true, 0, "WINDOWS", 0, "WINDOWS".length());
 
   /** Directory to store data. */
   private final File dataDirectory;
--- a/google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java
@@ -36,6 +36,7 @@ import java.nio.file.attribute.FileOwnerAttributeView;
 import java.nio.file.attribute.PosixFilePermission;
 import java.nio.file.attribute.UserPrincipal;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Set;
 import java.util.logging.Logger;
 
@@ -54,7 +55,7 @@ public class FileDataStoreFactory extends AbstractDataStoreFactory {
   private static final Logger LOGGER = Logger.getLogger(FileDataStoreFactory.class.getName());
 
   private static final boolean IS_WINDOWS = StandardSystemProperty.OS_NAME.value()
-      .regionMatches(true, 0, "WINDOWS", 0, "WINDOWS".length());
+      .toLowerCase(Locale.ENGLISH).startsWith("windows");
 
   /** Directory to store data. */
   private final File dataDirectory;
--- a/google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java
@@ -32,6 +32,7 @@ import java.nio.file.attribute.AclEntry;
 import java.nio.file.attribute.AclEntryPermission;
 import java.nio.file.attribute.AclEntryType;
 import java.nio.file.attribute.AclFileAttributeView;
+import java.nio.file.attribute.FileOwnerAttributeView;
 import java.nio.file.attribute.PosixFilePermission;
 import java.nio.file.attribute.UserPrincipal;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java
```diff
@@ -53,7 +53,7 @@
   private static final Logger LOGGER = Logger.getLogger(FileDataStoreFactory.class.getName());
 
   private static final boolean IS_WINDOWS = StandardSystemProperty.OS_NAME.value()
-      .startsWith("WINDOWS");
+      .regionMatches(true, 0, "WINDOWS", 0, "WINDOWS".length());
```
elharo: This appears to depend on the default character set.

diesieben07: We are just dealing with strings here, they are always in utf-16. Do you mean default locale? If so, I don't think so. The Javadoc for `regionMatches` does not mention anything about locale and the source code also does not do anything locale dependent.

diesieben07: As for testing...
The PR that introduced this check mentions waiting for a windows test machine, which was then apparently addressed in a different PR. I am not familiar with the automated testing setup for this library, but through searching the web I have not been able to find an instance where `os.name` _ever_ contained `"WINDOWS"` (all uppercase). So I am not sure how the initial PR even passed testing on any kind of Windows system.

diesieben07: See: https://github.com/googleapis/google-http-java-client/pull/557

diesieben07:
Turns out the implementation in `setPermissionsToOwnerOnlyWindows` did not actually work (as indicated by the [test failures that occurred on Windows](https://source.cloud.google.com/results/invocations/0991a4a3-62a4-4d8e-b743-83ec350f58ef/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fgoogle-http-java-client%2Fpresubmit%2Fjava8-win/log).
To me this means that when the original PR happened (https://github.com/googleapis/google-http-java-client/pull/557) it just never got tested, because the `IS_WINDOWS` test never worked. This caused the Windows test runs to also use `setPermissionsToOwnerOnly`, which does not _fail_ the tests, it simply logs a warning.

I've fixed obtaining the owning `UserPrinicipal`, which should also fix the tests on windows.

google-http-client/src/main/java/com/google/api/client/util/store/FileDataStoreFactory.java
elharo: Case insensitive comparison is fundamentally locale dependent. E.g. I is not the upper case of i in Turkish locales. UTF-16 is a character set, not a locale.

diesieben07: > UTF-16 is a character set, not a locale.
Yes. Which is why your original message saying "default character set" confused me.

> Case insensitive comparison is fundamentally locale dependent. E.g. I is not the upper case of i in Turkish locales. 
I know. `regionMatches` does not take this into account. `regionMatches` uses `Character.toLowerCase`, which specifically mentions:
> {@code String} case mapping methods can perform locale-sensitive mappings, context-sensitive mappings, and 1:M character mappings, whereas the {@code Character} case mapping methods cannot.

If you prefer i'll change it to a `toLowerCase(Locale.ENGLISH)` version though.

elharo: Yes, that would be good. I still need to think about how to test this.

diesieben07:
I've changed the code to use `toLowerCase(Locale.ENGLISH)` instead to be more explicit about localization rules.

Not sure why that test failed, it seems to be "unrelated flakyness", like you said.

diesieben07:
> Also, am I understanding you that the first change in line 56 by itself caused existing tests to fail, and then the additional change in line 160 caused the tests to pass again?

Correct. The change in line 56 changes Windows to actually use `setPermissionsToOwnerOnlyWindows`. The original implementation of that method was broken and threw an exception, causing test failures on Windows to now happen.
The change in line 160 fixes `setPermissionsToOwnerOnlyWindows` to actually work and no longer throw exceptions.

cryptearth:
As I'm not used to Github bare with me if this is inappropriate:
Although I understand the reason behind this check and change it's not common in the Windows world to do such specific permission stuff. Usually on a Windows system the NTFS does a good job with its ACLs to "guard" a users home directory from access by other non-Administrators (same as on Unix: if you're Administrator (root) permissions doesn't apply to you) and any other directory is usual on a somewhat semi-read-only. This all started when all that privilege stuff was introduced to normal people when XP was released, wich was based on Win2k and its NT based security model. The "normal user" didn't wanted to struggle about permissions and rights/access management - so Microsoft itself always had it rather loosely and since UAC came with Vista it was, and still is, common practice to just modify ACLs so that just anyone has full access (pretty much like a 777 on Unix) as most stuff struggle with permissions is used by normal home users on a home version of Windows.
TLDR: I don't see that windows specific check and modification useful at all - aside from the fact that when it was implemented at first no Windows box was available for testing. This may will cause more issues in the future as the broken one just because it's done right now.

PR #957, opened 2020-01-22
add PercentEscaperTest

elharo:
@frankyn I can't find any evidence that my previous change had an effect here.


frankyn:
Followed-up through chat. You're right. I reverted your change locally and my tests still pass. The fix is already in master AFAICT.

PR #959, opened 2020-01-24
Deprecate plusForSpace

elharo:
@chingor13 @frankyn This should be fully backwards compatible at an an API and behavior level. However it does strongly warn developers not to encode spaces as plus. I'd prefer to disable that completely, but the last time we tried that it broke some downstream projects, so this is a more nudging approach to get clients to stop doing this.


frankyn:
@elharo could you verify which google projects are using this?

I'd prefer not having customers start seeing deprecation warnings that we didn't prepare for first.

elharo:
Most usages of this should be in our own libraries; e.g. in java-storage. This client is rarely called directly by third party code, so I don't expect a lot of deprecation warnings in customer builds.

PR #983, opened 2020-02-23
fix for #982

### LINKED ISSUE #982: calling getCurrentToken() twice in com.google.api.client.json.JsonParser
In com.google.api.client.json.JsonParser in parseValue(Field, Type, ArrayList, Object, CustomizeJsonParser, boolean) at about line 720 local member JsonToken token is declared and assigned with call to getCurrentToken(). Two lines below the switch() calls getCurrentToken() again instead of use just declared and assigned token local member. Should be refactored so the switch uses the just declared and assigned member.

cryptearth:
changed switch parameter in parseValue from calling getCurrentToken() again to just declared and assigned local member "token" two lines above

Fixes #982


cryptearth:
@googlebot I signed it!

PR #986, opened 2020-03-02
add http.status_code attribute to all Spans that have a lowLevelHttpResponse 

ecourreges-orange:
This one is fairly straight forward:
It adds the official OpenTelemetry http.status_code attribute which is [mandatory](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md
), was missing and is very useful to know what happened to one's request.

Signed-off-by: CI-Bot for Emmanuel Courreges <emmanuel.courreges@orange.com>

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
@@ -1012,6 +1012,7 @@ public final class HttpRequest {
         LowLevelHttpResponse lowLevelHttpResponse = lowLevelHttpRequest.execute();
         if (lowLevelHttpResponse != null) {
           OpenCensusUtils.recordReceivedMessageEvent(span, lowLevelHttpResponse.getContentLength());
+          span.putAttribute(HttpTraceAttributeConstants.HTTP_STATUS_CODE, AttributeValue.longAttributeValue(lowLevelHttpResponse.getStatusCode()));
         }
         // Flag used to indicate if an exception is thrown before the response is constructed.
         boolean responseConstructed = false;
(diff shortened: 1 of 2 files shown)


ecourreges-orange:
> This library uses OpenCensus, not OpenTelemetry.
I know but OpenTelemetry will be adopted eventually
> 
> Do you happen to know what the status of those two is? We really need to figure that out.
As you can see [here](https://github.com/open-telemetry/opentelemetry-java), it is still a work in progress, so if I were you, I would wait until it is stable and production ready to migrate to it, but it will be fairly straightforward because there is an OpenCensus bridge, and also the interfaces are very similar.

google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
```diff
@@ -1012,6 +1012,7 @@ public HttpResponse execute() throws IOException {
         LowLevelHttpResponse lowLevelHttpResponse = lowLevelHttpRequest.execute();
         if (lowLevelHttpResponse != null) {
           OpenCensusUtils.recordReceivedMessageEvent(span, lowLevelHttpResponse.getContentLength());
+          span.putAttribute(HttpTraceAttributeConstants.HTTP_STATUS_CODE, AttributeValue.longAttributeValue(lowLevelHttpResponse.getStatusCode()));
```
elharo: Why is this a long instead of an int? Status codes don't go past 5XX

ecourreges-orange: Because there is only long for integer attribute types in OpenCensus: https://javadoc.io/doc/io.opencensus/opencensus-api/0.12.2/io/opencensus/trace/AttributeValue.html

ecourreges-orange:
@googlebot I signed it!

ecourreges-orange:
@googlebot I signed it!

elharo (approved):

PR #990, opened 2020-03-06
fix: Correctly handling chunked response streams with gzip

### LINKED ISSUE #367: chunked encoding + GZIP prevents keep-alive

hiranya911:
Fixes #367 (see my last comment on that issue)

The fix done in #840 is incorrect if not inadequate. It doesn't properly cleanup the last chunk of a chunked response when the response is also gzipped. This in turns prevents proper connection reuse. 

This PR addresses this issue.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -354,10 +354,10 @@ public final class HttpResponse {
             String oontentencoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
             if (CONTENT_ENCODING_GZIP.equals(oontentencoding) || CONTENT_ENCODING_XGZIP.equals(oontentencoding)) {
               // Wrap the original stream in a ConsumingInputStream before passing it to
-              // GZIPInputStream.
-              // GZIPInputStream tends to leave content in the original stream (it will almost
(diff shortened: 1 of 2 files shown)


elharo:
@codyoss Can you take a look at this and see what you think? Thanks.

hiranya911:
Hi Folks. Please let me know I need to do anything before this can be merged.

PR #1002, opened 2020-03-14
fix: include request method and URL into HttpResponseException message

medb:
This will allow to simplify debugging of failed requests, especially requests that are issued by [`MediaHttpUploader`](https://github.com/googleapis/google-api-java-client/blob/master/google-api-client/src/main/java/com/google/api/client/googleapis/media/MediaHttpUploader.java) (not created directly by library clients).

For example, it's useful to have an access to `upload_id` for failed resumable upload requests.

### DIFF
--- a/google-http-client/src/test/java/com/google/api/client/http/HttpResponseExceptionTest.java
+++ b/google-http-client/src/test/java/com/google/api/client/http/HttpResponseExceptionTest.java
@@ -44,12 +44,12 @@ public class HttpResponseExceptionTest extends TestCase {
     HttpRequest request = transport.createRequestFactory().buildGetRequest(SIMPLE_GENERIC_URL);
     HttpResponse response = request.execute();
     HttpHeaders headers = response.getHeaders();
-    HttpResponseException e = new HttpResponseException(response);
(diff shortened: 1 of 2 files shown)


google-http-client/src/test/java/com/google/api/client/http/HttpResponseExceptionTest.java
```diff
@@ -37,12 +41,11 @@
 
   public void testConstructor() throws Exception {
     HttpTransport transport = new MockHttpTransport();
-    HttpRequest request =
-        transport.createRequestFactory().buildGetRequest(HttpTesting.SIMPLE_GENERIC_URL);
+    HttpRequest request = transport.createRequestFactory().buildGetRequest(SIMPLE_GENERIC_URL);
     HttpResponse response = request.execute();
     HttpHeaders headers = response.getHeaders();
     HttpResponseException e = new HttpResponseException(response);
```
elharo: e --> exception (per Google style don't abbreviate variable names)

medb: Done.

PR #1006, opened 2020-03-20
fix: restore the thread's interrupted status after catching InterruptedException (#1005)

### LINKED ISSUE #1005: HttpBackOffIOExceptionHandler and HttpBackOffUnsuccessfulResponseHandler do not restore the interrupted status after catching InterruptionException
I found this by accident while looking at a different unexpected behavior within my system.

https://github.com/googleapis/google-http-java-client/blob/8760ff0e64ab47abba0f911d29d1afcb19dcfa34/google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java#L100

While handling IOException, the HttpBackOffIOExceptionHandler does not rethrow an exception and also does not reset the thread interrupted bit after catching InterruptedException.

See article with explanation: https://dzone.com/articles/why-do-we-need-threadcurrentthreadinterrupt-in-int

#### Environment details

1. Specify the API: GCS objects list
2. OS type and version: Debian
3. Java version: Java 8
4. google-http-client version(s): 1.34.2

micheldavid:
Fixes #1005

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java
@@ -97,11 +97,7 @@ public class HttpBackOffIOExceptionHandler implements HttpIOExceptionHandler {
     try {
       return BackOffUtils.next(sleeper, backOff);
     } catch (InterruptedException exception) {
-<<<<<<< HEAD
-      // Catching InterruptedException clears the thread interrupted bit.
-=======
       // Mark thread as interrupted since we cannot throw InterruptedException here.
->>>>>>> e66dee68... fix: reset the thread's interrupt bit after catching InterruptedException (#1005)
       Thread.currentThread().interrupt();
       return false;
     }
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java
@@ -97,7 +97,11 @@ public class HttpBackOffIOExceptionHandler implements HttpIOExceptionHandler {
     try {
       return BackOffUtils.next(sleeper, backOff);
     } catch (InterruptedException exception) {
+<<<<<<< HEAD
       // Catching InterruptedException clears the thread interrupted bit.
+=======
+      // Mark thread as interrupted since we cannot throw InterruptedException here.
+>>>>>>> e66dee68... fix: reset the thread's interrupt bit after catching InterruptedException (#1005)
       Thread.currentThread().interrupt();
       return false;
     }
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java
@@ -97,6 +97,8 @@ public class HttpBackOffIOExceptionHandler implements HttpIOExceptionHandler {
     try {
       return BackOffUtils.next(sleeper, backOff);
     } catch (InterruptedException exception) {
+      // Catching InterruptedException clears the thread interrupted bit.
+      Thread.currentThread().interrupt();
       return false;
     }
   }
--- a/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpTransport.java
+++ b/google-http-client-apache-v2/src/main/java/com/google/api/client/http/apache/v2/ApacheHttpTransport.java
@@ -38,19 +38,15 @@ import org.apache.http.impl.conn.SystemDefaultRoutePlanner;
 /**
  * Thread-safe HTTP transport based on the Apache HTTP Client library.
  *
- * <p>
- * Implementation is thread-safe, as long as any parameter modification to the
- * {@link #getHttpClient() Apache HTTP Client} is only done at initialization time. For maximum
- * efficiency, applications should use a single globally-shared instance of the HTTP transport.
- * </p>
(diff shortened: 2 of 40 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java
```diff
@@ -97,6 +97,8 @@ public boolean handleIOException(HttpRequest request, boolean supportsRetry) thr
     try {
       return BackOffUtils.next(sleeper, backOff);
     } catch (InterruptedException exception) {
+      // Catching InterruptedException clears the thread interrupted bit.
```
elharo: I could well be wrong here. I usually am when it comes to Thread arcana, but reading that article and the docs, this only seems necessary when you call `interrupted()`, not when you simply catch InterruptedException. That is, this code does not clear the interrupted bit. Am I missing something?

chanseokoh: This may well be the right thing to do.

Unless this code has control, authority, and context or institutes a policy so that eventually it can make the current interrupted thread die (or decide not to terminate the thread as its normal operation)–which I don't think is the case as a library, or unless this code re-throws `InterruptedException`, or unless it is really correct that this code is designed to consume thread interruption and exit by returning false, I think it should restore the interrupted bit, so that other parts of the code (someone higher in the call stack) can be made aware that this thread is interrupted and act accordingly.

https://stackoverflow.com/questions/2523721/why-do-interruptedexceptions-clear-a-threads-interrupted-status
https://stackoverflow.com/questions/4906799/why-invoke-thread-currentthread-interrupt-in-a-catch-interruptexception-block

But I think the comment `// Catching ... clears the thread interrupted bit.` is unnecessary.

chanseokoh: And I actually think the comment "Catching InterruptedException clears the thread interrupted bit" is incorrect and misleading. It's not the act of catching it. Rather, I believe it is just that the interrupted bit is already cleared (and should by convention) when `InterruptedException` is thrown. From https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html,

> By convention, any method that exits by throwing an InterruptedException clears interrupt status when it does so.

elharo: It would help if this had a test demonstrating both that the bug is real and that this patch fixes it.

micheldavid: Thanks for the quick response and thoughtful discussion.

Sorry about the delay. I havent had spare cycles to work on this recently...
I will write a test, but note that it is difficult to predict threading behavior so the test may be flaky at times, depending on cpu speed and available resources.

elharo: Testing thread issues is extremely tricky. A flaky test can be worse than no test. This might need an integration test that doesn't run as part of the normal suite or some other magic.

elharo:
This might help with testing:

https://github.com/awaitility/awaitility

micheldavid:
Rebased and added tests. I also re-interrupted the thread in `HttpBackOffUnsuccessfulResponseHandler`.

I noticed [HttpRequest is swallowing an InterruptedException too](https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java#L1071), but that code depends on [BackOffPolicy which javadoc says](https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/BackOffPolicy.java#L26) is `scheduled to be removed in 1.18` so I did not bother fixing it.

A bit more literature around interrupted exceptions: go/interrupted-exception (internal) and http://www.ibm.com/developerworks/java/library/j-jtp05236/index.html

google-http-client/src/main/java/com/google/api/client/http/HttpBackOffIOExceptionHandler.java
```diff
@@ -97,6 +97,12 @@ public boolean handleIOException(HttpRequest request, boolean supportsRetry) thr
     try {
       return BackOffUtils.next(sleeper, backOff);
     } catch (InterruptedException exception) {
+<<<<<<< HEAD
```
elharo, line 100: bad merge

micheldavid: oops... cant believe this got in here without me noticing.
fixed. ptal.

micheldavid:
Thanks for the quick turn around. I see a lot of files I did not change are failing the code format lint pass. I did change some of them though. Should I fix it with "mvn com.coveo:fmt-maven-plugin:format"? Maybe fix only the ones I changed?

I also do not have write access, so I will need someone to merge this in.

elharo:
yes, please run the plugin

micheldavid:
Now the linter is failing on the `google-http-client-bom` project, which doesnt seem to have any code...

From https://source.cloud.google.com/results/invocations/7a0f7d52-8162-4c7b-b711-f7c90e073a42/targets/cloud-devrel%2Fclient-libraries%2Fjava%2Fgoogle-http-java-client%2Fpresubmit%2Flint/log

Failed to execute goal com.coveo:fmt-maven-plugin:2.10:check (default-cli) on project google-http-client-bom

Is this normal? The same seems to happen in master.
I feel like these formatting issues should be resolved in a separate pull request.

chingor13:
Strange, I don't recall us running the code formatter on this repo. Looking into it

chingor13:
We can fix the linting issue separately.

micheldavid:
Thanks Jeff!

Issue #926, opened 2019-12-24
Api 19  Caused by: java.lang.NoClassDefFoundError: java.nio.charset.StandardCharsets


Zevsus:
#### Environment details

2. OS type and version: Android api 19
3. Java version: 8
4. google-http-client version(s):  implementation 'com.google.http-client:google-http-client-gson:1.34.0'

#### Code example
private static List<File> exeQ(String q) throws IOException {
        String nextPageToken = "";
        List<File> res = new ArrayList<>();
        do {
            final FileList list = googleDriveService.files().list().setSpaces("drive").setQ(q).setPageToken(nextPageToken).setFields("nextPageToken, files(*)").setPageSize(100).setOrderBy("modifiedTime").execute();
            nextPageToken = list.getNextPageToken();
            res.addAll(list.getFiles());
        } while (nextPageToken != null);
        return res;
    }

#### Stack trace
```
     Caused by: java.lang.NoClassDefFoundError: java.nio.charset.StandardCharsets
        at com.google.api.client.util.escape.CharEscapers.decodeUriPath(CharEscapers.java:110)
        at com.google.api.client.http.GenericUrl.toPathParts(GenericUrl.java:594)
        at com.google.api.client.http.GenericUrl.<init>(GenericUrl.java:200)
        at com.google.api.client.http.GenericUrl.<init>(GenericUrl.java:177)
        at com.google.api.client.http.GenericUrl.<init>(GenericUrl.java:126)
        at com.google.api.client.http.GenericUrl.<init>(GenericUrl.java:109)
        at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.buildHttpRequestUrl(AbstractGoogleClientRequest.java:388)
        at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.buildHttpRequest(AbstractGoogleClientRequest.java:423)
        at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:542)
        at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.executeUnparsed(AbstractGoogleClientRequest.java:475)
        at com.google.api.client.googleapis.services.AbstractGoogleClientRequest.execute(AbstractGoogleClientRequest.java:592)
        (GFile.java:192)
```
Thanks!

codyoss:
From what I can tell StandardCharsets was added in with api level 19, so this should work. 
[...]

PR #1007, opened 2020-03-20
docs: require Android 4.4

elharo:
@chingor13 @BenWhitehead We now require API 19 or higher which means Android 4.4 or later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@ content. The JSON and XML libraries are also fully pluggable, and they include s
 
 The library supports the following Java environments:
 
-- Java 7 (or higher)
-- Android 4.0 (Ice Cream Sandwich) (or higher)
+- Java 7 or higher
+- Android 4.4 (Kit Kat)
 - GoogleAppEngine Google App Engine
 
 The following related projects are built on the Google HTTP Client Library for Java:


chingor13:
Alternatively, we could provide the old implementation that was incompatible with windows in the google-http-client-android extensions artifact. This is only update we've taken that bumped the Android requirement.

elharo:
I'm not sure I follow your comment. Perhaps file a separate issue for that?

Meanwhile, the released version of this library does require Android 4.4 or later and the docs should reflect that.

chingor13:
Only one class requires Android 4.4 (FileDatastoreFactory). If it's not a compile time issue, we can provide an alternative implementation (the original one that does not work on windows) in the google-http-client-android extensions artifact. If it is a compile time issue, we can protect with reflection.

elharo:
GenericUrl also requires it. See https://github.com/googleapis/google-http-java-client/issues/926

SanoKriss:
> Only one class requires Android 4.4 (FileDatastoreFactory). If it's not a compile time issue, we can provide an alternative implementation (the original one that does not work on windows) in the google-http-client-android extensions artifact. If it is a compile time issue, we can protect with reflection.

Yes please, make a fork compatible with API 17 Android, thanks

PR #1033, opened 2020-04-24
feat: add logic for verifying ES256 JsonWebSignatures

chingor13:
Adds logic for verifying signatures that use NIST P-256 (ES256). It is up to the caller to provide the correct public key for verification.

Note that the difference between the ES256 and RS256 verification process is that ES256 requires AS1 (DER) notation for the signature so we do a limited conversion here before providing the signature to the signing algorithm.

Logic will be added to `google-auth-library` for handling Google's well-known, published public keys.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
+++ b/google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
@@ -23,6 +23,7 @@ import com.google.api.client.util.SecurityUtils;
 import com.google.api.client.util.StringUtils;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.math.BigInteger;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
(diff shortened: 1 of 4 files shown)


google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
```diff
@@ -574,4 +574,35 @@ public static String signUsingRsaSha256(
             SecurityUtils.getSha256WithRsaSignatureAlgorithm(), privateKey, contentBytes);
     return content + "." + Base64.encodeBase64URLSafeString(signature);
   }
+
+  static class DerEncoder {
+    private static byte DER_TAG_SIGNATURE_OBJECT = 0x30;
+    private static byte DER_TAG_ASN1_INTEGER = 0x02;
+
+    static byte[] encode(byte[] signature) {
```
elharo, line 582: why the fully qualified name?

google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
```diff
@@ -574,4 +574,35 @@ public static String signUsingRsaSha256(
             SecurityUtils.getSha256WithRsaSignatureAlgorithm(), privateKey, contentBytes);
     return content + "." + Base64.encodeBase64URLSafeString(signature);
   }
+
+  static class DerEncoder {
```
elharo, line 578: perhaps make this an outer class

google-http-client/src/main/java/com/google/api/client/json/webtoken/JsonWebSignature.java
chingor13: Oops, this was migrated from WIP on the `google-auth-library` which used guava's `Preconditions`. `google-http-client` has legacy wrappers for guava's `Preconditions` (with the same name in the current package) from when it was shaded and included in this jar.

I'll switch to the wrapper to maintain consistency with the rest of the library's usage of Preconditions.

PR #1046, opened 2020-05-21
fix: make google-http-client.properties file shading friendly

medb:
Fixes an issue when there are 2 versions of this library in a class path - 1) shaded/relocated library of `x.x.x` version inside a "jar-with-dependencies" and 2) non-shaded library of `y.y.y` version.

Because both of these libraries will load non-shaded `google-http-client.properties` file they will use the same `y.y.y` version even though shaded library has a `x.x.x` version.


BenWhitehead:
@medb Can you describe why this change is needed? The properties file should be able to be loaded from an uber-jar or other shaded jar just fine without modification.

I have a local project on my machine I used to execute a few commands against Firestore (which is a downstream consumer of this library). This project is packaged as a "jar-with-dependencies" file. `HttpRequest` is able to load the properties files successfully as-is.

```
❱ java -jar target/cleanup-0.1.1-SNAPSHOT-jar-with-dependencies.jar 
HttpRequest.VERSION = 1.34.2
```

From this code
```java
public static void main(String[] args) {
  System.out.println("HttpRequest.VERSION = " + com.google.api.client.http.HttpRequest.VERSION);
}
```

medb:
The issue is that if there are 2 versions of this library in a class path - 1) shaded library of `x.x.x` version inside a "jar-with-dependencies" and 2) non-shaded library of `y.y.y` version.

Because both of these libraries will load non-shaded `google-http-client.properties` they will use the same `y.y.y` version even though shaded library has a `x.x.x` version.

BenWhitehead:
@chingor13 What are you thoughts here on moving the properties file? We might need to move it for the couple other libraries we introduced this same pattern in.

chingor13 (approved):

PR #1070, opened 2020-06-26
feat: add Android 19 compatible FileDataStoreFactory implementation

### LINKED ISSUE #906: google-http-client-android 1.31+ Posix crash
> com.google.api.client.util.store.FileDataStoreFactory.setPermissionsToOwnerOnly

uses `PosixFilePermission`
This class was added only in Android API 26. Running on lower API devices results in:

> Fatal Exception: java.lang.NoClassDefFoundError
Failed resolution of: Ljava/nio/file/attribute/PosixFilePermission;
com.google.api.client.util.store.FileDataStoreFactory.setPermissionsToOwnerOnly

Crash happens when trying to store OAuth credentials and doing something trivial like

> new FileDataStoreFactory(filesDir)

Maybe since the package releases as "Android-compatible" this should be documented?

chingor13:
Fixes #906

* Adds an Android compatible FileDataStoreFactory implementation (version prior to switching to NIO) in the google-http-client-android extensions artifact
* Adds a note in the core FileDataStoreFactory to use the new version on Android API level < 26
* Adds an animal-sniffer check for Android API level 19 (Ignores the NIO Files API usage).

### DIFF
--- a/google-http-client-android/src/main/java/com/google/api/client/extensions/android/util/store/FileDataStoreFactory.java
+++ b/google-http-client-android/src/main/java/com/google/api/client/extensions/android/util/store/FileDataStoreFactory.java
@@ -121,38 +121,22 @@ public class FileDataStoreFactory extends AbstractDataStoreFactory {
    * executed by the file's owner.
    *
    * @param file the file's permissions to modify
-   * @throws IOException
    */
-  static void setPermissionsToOwnerOnly(File file) throws IOException {
+  static void setPermissionsToOwnerOnly(File file) {
     // Disable access by other users if O/S allows it and set file permissions to readable and
-    // writable by user. Use reflection since JDK 1.5 will not have these methods
+    // writable by user.
     try {
-      Method setReadable = File.class.getMethod("setReadable", boolean.class, boolean.class);
-      Method setWritable = File.class.getMethod("setWritable", boolean.class, boolean.class);
-      Method setExecutable = File.class.getMethod("setExecutable", boolean.class, boolean.class);
-      if (!(Boolean) setReadable.invoke(file, false, false)
-          || !(Boolean) setWritable.invoke(file, false, false)
-          || !(Boolean) setExecutable.invoke(file, false, false)) {
+      if (!file.setReadable(false, false)
+          || !file.setWritable(false, false)
+          || !file.setExecutable(false, false)) {
         LOGGER.warning("unable to change permissions for everybody: " + file);
       }
-      if (!(Boolean) setReadable.invoke(file, true, true)
-          || !(Boolean) setWritable.invoke(file, true, true)
-          || !(Boolean) setExecutable.invoke(file, true, true)) {
+      if (!file.setReadable(true, true)
+          || !file.setWritable(true, true)
+          || !file.setExecutable(true, true)) {
         LOGGER.warning("unable to change permissions for owner: " + file);
       }
-    } catch (InvocationTargetException exception) {
-      Throwable cause = exception.getCause();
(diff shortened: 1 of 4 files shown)


google-http-client-android/src/main/java/com/google/api/client/extensions/android/util/store/FileDataStoreFactory.java
```diff
@@ -0,0 +1,161 @@
...
+    } catch (InvocationTargetException exception) {
+      Throwable cause = exception.getCause();
+      Throwables.propagateIfPossible(cause, IOException.class);
+      // shouldn't reach this point, but just in case...
+      throw new RuntimeException(cause);
+    } catch (NoSuchMethodException exception) {
+      LOGGER.warning(
+          "Unable to set permissions for "
+              + file
+              + ", likely because you are running a version of Java prior to 1.6");
```
elharo, line 152: Is this the right error message here?

chingor13: Probably not, this was the version from google-http-client (not restricted to android). I'll see if I can come up with a better message.

chingor13: Removed the reflection as the methods we were previously protecting were added in [API level 9](https://developer.android.com/reference/java/io/File#setReadable(boolean,%2520boolean)).

google-http-client-android/src/main/java/com/google/api/client/extensions/android/util/store/FileDataStoreFactory.java
```diff
@@ -0,0 +1,142 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.logging.Logger;
+
+/**
+ * Thread-safe file implementation of a credential store.
+ *
+ * <p>For security purposes, the file's permissions are set to be accessible only by the file's
+ * owner. Note that Java 1.5 does not support manipulating file permissions, and must be done
```
elharo, line 34: no need to mention Java 1.5 any more

google-http-client-android/src/main/java/com/google/api/client/extensions/android/util/store/FileDataStoreFactory.java
```diff
@@ -0,0 +1,142 @@
...
+
+  /** Directory to store data. */
+  private final File dataDirectory;
+
+  /** @param dataDirectory data directory */
+  public FileDataStoreFactory(File dataDirectory) throws IOException {
+    dataDirectory = dataDirectory.getCanonicalFile();
+    this.dataDirectory = dataDirectory;
+    // error if it is a symbolic link
+    if (IOUtils.isSymbolicLink(dataDirectory)) {
```
elharo, line 55: the object has been modified before the exception is thrown

google-http-client-android/src/main/java/com/google/api/client/extensions/android/util/store/FileDataStoreFactory.java
chingor13: Done, removed.

google-http-client-android/src/main/java/com/google/api/client/extensions/android/util/store/FileDataStoreFactory.java
chingor13: Done. Moved the member assignment after the validations here and in the mainline implementation.

PR #1100, opened 2020-08-28
Feature #1098: Improve UrlEncodedContent class

### LINKED ISSUE #1098: UrlEncodedContent class should use URI path encoding
UrlEncodedContent class use a deprecated URL encoder function. For another feature request in Google-auth-library-java (I will paste the link when the issue will be created), I need to use URI path encoding (especially for not encoding the + when I define a list of scopes in the auth library).

guillaumeblaquiere:
Fixes #1098 
The legacy behavior is kept. Test has been updated

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
@@ -42,18 +42,33 @@ import java.util.Map;
  *
  * <p>Implementation is not thread-safe.
  *
- * @since 1.0
  * @author Yaniv Inbar
+ * @since 1.0
  */
 public class UrlEncodedContent extends AbstractHttpContent {
 
   /** Key name/value data. */
   private Object data;
 
+  /** Use URI Path encoder flag. False by default (use legacy and deprecated escapeUri) */
+  private Boolean useEscapeUriPathEncoding;
+
   /** @param data key name/value data */
   public UrlEncodedContent(Object data) {
     super(UrlEncodedParser.MEDIA_TYPE);
     setData(data);
+    this.useEscapeUriPathEncoding = false;
+  }
+
+  /**
+   * @param data key name/value data
+   * @param useEscapeUriPathEncoding replace the legacy (and deprecated) URI encoder, by the
+   *     EncodedPathUri encoder
+   */
+  public UrlEncodedContent(Object data, Boolean useEscapeUriPathEncoding) {
+    super(UrlEncodedParser.MEDIA_TYPE);
+    setData(data);
+    this.useEscapeUriPathEncoding = useEscapeUriPathEncoding;
   }
 
   public void writeTo(OutputStream out) throws IOException {
@@ -66,10 +81,10 @@ public class UrlEncodedContent extends AbstractHttpContent {
         Class<? extends Object> valueClass = value.getClass();
         if (value instanceof Iterable<?> || valueClass.isArray()) {
           for (Object repeatedValue : Types.iterableOf(value)) {
-            first = appendParam(first, writer, name, repeatedValue);
+            first = appendParam(first, writer, name, repeatedValue, this.useEscapeUriPathEncoding);
           }
         } else {
-          first = appendParam(first, writer, name, value);
+          first = appendParam(first, writer, name, value, this.useEscapeUriPathEncoding);
         }
       }
     }
@@ -125,7 +140,8 @@ public class UrlEncodedContent extends AbstractHttpContent {
     return result;
   }
 
-  private static boolean appendParam(boolean first, Writer writer, String name, Object value)
+  private static boolean appendParam(
+      boolean first, Writer writer, String name, Object value, boolean useEscapePathEncoding)
       throws IOException {
     // ignore nulls
     if (value == null || Data.isNull(value)) {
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
```diff
@@ -42,18 +42,33 @@
...
  * @author Yaniv Inbar
+ * @since 1.0
  */
 public class UrlEncodedContent extends AbstractHttpContent {
 
   /** Key name/value data. */
   private Object data;
 
+  /** Use URI Path encoder flag. False by default (use legacy and deprecated escapeUri) */
+  private Boolean useEscapeUriPathEncoding;
```
elharo, line 54: boolean

elharo, line 54: rename this please to something shorter without "use"

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
```diff
@@ -42,18 +42,33 @@
...
+
   /** @param data key name/value data */
   public UrlEncodedContent(Object data) {
     super(UrlEncodedParser.MEDIA_TYPE);
     setData(data);
+    this.useEscapeUriPathEncoding = false;
+  }
+
+  /**
+   * @param data key name/value data
```
elharo, line 64: the key is a name and a value? That's surprising.

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
```diff
@@ -42,18 +42,33 @@
...
+   * @param useEscapeUriPathEncoding replace the legacy (and deprecated) URI encoder, by the
```
elharo, line 65: describe this in terms of behavior rather than implementation. The user doesn't care which encoder is used, only how URIs are encoded.

elharo (changes requested):
I wish we could do this unconditionally but that's risky.

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
guillaumeblaquiere: Done

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
guillaumeblaquiere: Copy paste of the legacy comment (see line 56 of the file.) I kept the legacy signature and behavior, and I added mine with additional parameter. This one is unchanged (and I guess its description also!)

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
guillaumeblaquiere: I tried to improve it. Any feedback is welcomed!!

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
guillaumeblaquiere: I tried a new name. Let me know what do you think.

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
guillaumeblaquiere: Let me know if you want me to improve both or if keeping the existing description is ok.

elharo: The more I look at this the weirder it is. Why is this an object instead of a Map?

guillaumeblaquiere: I agree; it's strange and I don't know why to use an Object instead of a Map! However to ensure the compatibility with other app that depends on this library, I kept the existing signature.

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
```diff
@@ -53,25 +53,30 @@
...
+   *  @param data key name/value data
+   *  */
   public UrlEncodedContent(Object data) {
     super(UrlEncodedParser.MEDIA_TYPE);
     setData(data);
     this.uriPathEncodingFlag = false;
   }
 
   /**
+   * Initialize the UrlEncodedContent with our without the legacy and deprecated escapeUri encoder
```
elharo, line 67: our --> or?

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
```diff
@@ -53,25 +53,30 @@
...
    * @param data key name/value data
-   * @param useUriPathEncoding Escapes the string value so it can be safely included in URI path segments. For details on
-   *    * escaping URIs, see <a href="http://tools.ietf.org/html/rfc3986#section-2.4">RFC 3986 - section
-   *    * 2.4</a>
+   * @param useUriPathEncoding Escapes the string value so it can be safely included in URI path segments.
```
elharo, line 69: Escapes --> escapes

google-http-client/src/main/java/com/google/api/client/http/UrlEncodedContent.java
```diff
@@ -53,25 +53,30 @@
...
+   *     For details on escaping URIs, see <a href="http://tools.ietf.org/html/rfc3986#section-2.4">RFC 3986 -
+   *     section 2.4</a>
    */
   public UrlEncodedContent(Object data, Boolean useUriPathEncoding) {
```
elharo, line 73: Boolean --> boolean 
per Effective Java

elharo (approved):

PR #1174, opened 2020-11-04
feat: compile java 8 binaries

chingor13:
We are still checking Android compatibility via animal-sniffer


elharo (changes requested):
Is there a reason to do this now? This is the sort of change go/nojava7 is trying to avoid surprising customers with.

PR #1209, opened 2020-12-21
fix: use only Gson

### LINKED ISSUE #1208: Convert google-http-java-client/google-http-client-jackson2/ to be based on gson
Per go/json-madness

I *think* that despite the name of the package, the Jackson dependencies are purely internal so we should be able to swap this out without breaking anything.

elharo:
@chingor13 fixes #1208


chingor13:
I believe we decided to not do this and instead heavily suggest users to use gson.

PR #1221, opened 2021-01-13
deps: force commons-codec to latest version

elharo:
@stephaniewang526

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -101,6 +101,11 @@
   <dependencyManagement>
     <dependencies>
       <dependency>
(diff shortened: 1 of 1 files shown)


chingor13 (changes requested):
We need to also enable the flatten-maven-plugin or this will only locally use 1.15. The published pom will not include commons-codec and users would get 1.11 transitively via apache httpclient.

stephaniewang526:
We are managing this in shared-dependency as discussed: https://github.com/googleapis/java-shared-dependencies/pull/251.

Issue #1253, opened 2021-02-10
YouTube sample incorrectly formatted at head. 


elharo:
P1 since it breaks the CI checks

```
Error:  src/main/java/com/example/json/YouTubeSample.java:[8] (header) RegexpHeader: Line does not match expected header line of '^\W+\*?\W*http://www.apache.org/licenses/LICENSE-2.0$'.
Error:  src/main/java/com/example/json/YouTubeSample.java:[16,1] (whitespace) EmptyLineSeparator: 'package' should be separated from previous statement.
Error:  src/main/java/com/example/json/YouTubeSample.java:[21,1] (imports) CustomImportOrder: Extra separation in import group before 'java.io.IOException'
Error:  src/test/java/com/example/json/YouTubeSampleTest.java:[8] (header) RegexpHeader: Line does not match expected header line of '^\W+\*?\W*http://www.apache.org/licenses/LICENSE-2.0$'.
Error:  src/test/java/com/example/json/YouTubeSampleTest.java:[16,1] (whitespace) EmptyLineSeparator: 'package' should be separated from previous statement.
Error:  src/test/java/com/example/json/YouTubeSampleTest.java:[33,1] (imports) CustomImportOrder: Extra separation in import group before 'java.io.IOException'
Error:  src/test/java/com/example/json/YouTubeSampleTest.java:[33,1] (imports) CustomImportOrder: Wrong lexicographical order for 'java.io.IOException' import. Should be before 'org.junit.Test'.
Error:  src/test/java/com/example/json/YouTubeSampleTest.java:[34,1] (imports) CustomImportOrder: Wrong lexicographical order for 'java.io.InputStream' import. Should be before 'org.junit.Test'.
Error:  src/test/java/com/example/json/YouTubeSampleTest.java:[35,1] (imports) CustomImportOrder: Wrong lexicographical order for 'java.util.Map' import. Should be before 'org.junit.Test'.
Error:  src/test/java/com/example/json/YouTubeSampleTest.java:[37,1] (imports) CustomImportOrder: Import statement for 'org.junit.Assert.assertEquals' is in the wrong order. Should be in the 'STATIC' group, expecting not assigned imports on this line.
Error:  src/test/java/com/example/json/YouTubeSampleTest.java:[38,1] (imports) CustomImportOrder: Import statement for 'org.junit.Assert.assertNotNull' is in the wrong order. Should be in the 'STATIC' group, expecting not assigned imports on this line.
[...]

PR #1255, opened 2021-02-11
fix: fix buildRequest setUrl order

arithmetic1728:
`request.setUrl` should be called before `initializer.intialize(request)`, otherwise program may crash if request has a null uri.

For example, `HttpCredentialsAdapter` is a common implementation of `HttpRequestInitializer`. In its `initialize(request)` [implementation](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/http/HttpCredentialsAdapter.java#L91), it calls `Map<String, List<String>> credentialHeaders = credentials.getRequestMetadata(uri);`. 

Note if the credentials is a `ServiceAccountJWTAccessCredentials`, then [`getRequestMetadata(uri)`](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/ServiceAccountJwtAccessCredentials.java#L349) throws exception if both `uri` and `defaultAudience` are both null. Moreover, most times uri is the audience used for jwt token, it should be passed instead of being null all the time.


elharo (changes requested):
test?

arithmetic1728:
> test?

done.

google-http-client/src/test/java/com/google/api/client/http/HttpRequestFactoryTest.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2021 Google Inc.
```
elharo, line 2: No (c)
In fact, never c in parentheses anywhere. It is *not* a legal copyright symbol

arithmetic1728, line 2:
```suggestion
 * Copyright 2021 Google Inc.
```

arithmetic1728: done

arithmetic1728:
@elharo could you take another look? This fix will be needed for self signed jwt feature. Thanks!

google-http-client/src/test/java/com/google/api/client/http/HttpRequestFactoryTest.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2021 Google Inc.
```
elharo, line 2: LLC

elharo (commented):
tests are failing. Might not be related but should be fixed first

google-http-client/src/test/java/com/google/api/client/http/HttpRequestFactoryTest.java
arithmetic1728, line 2:
```suggestion
 * Copyright 2021 Google LLC.
```

arithmetic1728: done

arithmetic1728:
> tests are failing. Might not be related but should be fixed first

The failures are not related to my changes. They already failed in the recent submitted PRs. There were two failures: one is YouTube sample style check error, the other is Jackson 2 package javadoc error. 

I fixed the style check failure in the YouTube sample by running `mvn com.coveo:fmt-maven-plugin:format` and changing the license url (https->http). Now the error becomes enforcer errors, I don't have much experience with Java, not sure how to fix this kind of problem. I think probably we can check in this PR, I will file a bug to let the Java experts fix it.

```
[WARNING] Rule 2: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for com.google.errorprone:error_prone_annotations:2.3.4 paths to dependency are:
+-com.google.http-client:google-http-client:1.38.2-SNAPSHOT
  +-com.google.guava:guava:30.1-android
    +-com.google.errorprone:error_prone_annotations:2.3.4
and
+-com.google.http-client:google-http-client:1.38.2-SNAPSHOT
  +-com.google.guava:guava-testlib:30.1-android
    +-com.google.errorprone:error_prone_annotations:2.3.4
and
+-com.google.http-client:google-http-client:1.38.2-SNAPSHOT
  +-com.google.truth:truth:1.1.2
    +-com.google.errorprone:error_prone_annotations:2.5.1
]
```

elharo (commented):
Try just fixing the youtube samples in a separate PR. This is #1253 It's not related but it does need to be fixed before anything else can proceed.

chingor13 (commented):
This is potentially a breaking behavior change for someone if their request initializer is conditionally setting the url (which would be a slightly odd check because it would never be previously set).

The existing test cases do not guarantee this to be true, though.

google-http-client/src/test/java/com/google/api/client/http/HttpRequestFactoryTest.java
chingor13, line 2:
```suggestion
 * Copyright 2021 Google LLC
```

chingor13 (approved):

arithmetic1728:
> This is potentially a breaking behavior change for someone if their request initializer is conditionally setting the url (which would be a slightly odd check because it would never be previously set).
> 
> The existing test cases do not guarantee this to be true, though.

If their request initializer sets the url, then in the original code the url in the request will be overwritten, so it should be fine.
```
if (url != null) {
      request.setUrl(url);
    }
```

chingor13:
> > This is potentially a breaking behavior change for someone if their request initializer is conditionally setting the url (which would be a slightly odd check because it would never be previously set).
> > The existing test cases do not guarantee this to be true, though.
> 
> If their request initializer sets the url, then in the original code the url in the request will be overwritten, so it should be fine.
> 
> ```
> if (url != null) {
>       request.setUrl(url);
>     }
> ```

The edge case I'm thinking of is if their initializer conditionally sets the url if not set. In the current implementation, the initializer would have set it, but now if it's set earlier, their initializer wouldn't set it. Granted this is a strange usage of the initializer.

arithmetic1728:
@elharo can we merge this PR? Thanks!

elharo (approved):

PR #1268, opened 2021-02-18
deps: version manage error_prone_annotations to 2.5.1

chingor13:
Fixes the dependencies presubmits

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -127,6 +127,11 @@
         <scope>test</scope>
       </dependency>
       <dependency>
+        <groupId>com.google.errorprone</groupId>
+        <artifactId>error_prone_annotations</artifactId>
+        <version>2.5.1</version>
+      </dependency>
+      <dependency>
         <groupId>com.google.appengine</groupId>
         <artifactId>appengine-api-1.0-sdk</artifactId>
         <version>${project.appengine.version}</version>


pom.xml
```diff
@@ -126,6 +126,11 @@
         <version>1.1.2</version>
         <scope>test</scope>
       </dependency>
+      <dependency>
```
elharo, line 129: what's the root cause here? We shouldn't need this in this repo. The failures are weird as they seem flaky and perhaps dependent on java 8 vs 11. I haven't dug into this yt.

arithmetic1728: I think the problem is that the guava dependencies are using different versions of `error_prone_annotations` (two use 2.3.4 and another one uses 2.5.1), and this causes enforcer failure. This PR specifies the `error_prone_annotations` version in pom.xml so only this version will be used.

```
[WARNING] Rule 2: org.apache.maven.plugins.enforcer.RequireUpperBoundDeps failed with message:
Failed while enforcing RequireUpperBoundDeps. The error(s) are [
Require upper bound dependencies error for com.google.errorprone:error_prone_annotations:2.3.4 paths to dependency are:
+-com.google.http-client:google-http-client:1.38.2-SNAPSHOT
  +-com.google.guava:guava:30.1-android
    +-com.google.errorprone:error_prone_annotations:2.3.4
and
+-com.google.http-client:google-http-client:1.38.2-SNAPSHOT
  +-com.google.guava:guava-testlib:30.1-android
    +-com.google.errorprone:error_prone_annotations:2.3.4
and
+-com.google.http-client:google-http-client:1.38.2-SNAPSHOT
  +-com.google.truth:truth:1.1.2
    +-com.google.errorprone:error_prone_annotations:2.5.1
]
```

elharo: I sent a cl to update Guava, though it will be some time before that releases.

What I don't understand is why this problem is flaky. Sometimes I see it. Sometimes I don't. It feels like it should be always or never.

elharo (commented):
Can you merge in master? I think the jackson javadoc issue should be fixed now

elharo (approved):

suztomo:
Once this is merged, I'll have these GitHub Actions checks as required checks.
https://github.com/googleapis/google-http-java-client/pull/1273

PR #1305, opened 2021-03-02
fix: default application/json charset to utf-8

### LINKED ISSUE #1102: the default charset incorrect when the content-type is application/json
## Environment details

1.  HTTP Requests with the "content-type: applications/json" response
2. OS type and version: All
3. Java version: All
4. google-http-client version(s): All

#### Steps to reproduce

  1. A Server response "content-type: application/json" with some special characters like Chinese.
  2. From the https://tools.ietf.org/html/rfc4627, 3.Encoding "JSON text SHALL be encoded in Unicode.  The default encoding is  UTF-8."
  3. But in google-http-java-client/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java, it use the charset Charsets.ISO_8859_1.

Thanks!

chingor13:
Fixes #1102

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -18,7 +18,6 @@ import com.google.api.client.util.IOUtils;
 import com.google.api.client.util.LoggingInputStream;
 import com.google.api.client.util.Preconditions;
 import com.google.api.client.util.StringUtils;
-import com.google.common.annotations.VisibleForTesting;
 
 import java.io.ByteArrayOutputStream;
 import java.io.EOFException;
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -18,7 +18,6 @@ import com.google.api.client.util.IOUtils;
 import com.google.api.client.util.LoggingInputStream;
 import com.google.api.client.util.Preconditions;
 import com.google.api.client.util.StringUtils;
-
 import java.io.ByteArrayOutputStream;
 import java.io.EOFException;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -512,8 +514,18 @@ public String parseAsString() throws IOException {
...
+    if (mediaType != null) {
+      // use specified charset parameter from content/type header if available
+      if (mediaType.getCharsetParameter() != null) {
+        return mediaType.getCharsetParameter();
+      }
+      // fallback to well-known charsets
+      if (mediaType.getType().equals("application") && mediaType.getSubType().equals("json")) {
+        // https://tools.ietf.org/html/rfc4627 - JSON must be encoded with UTF-8
+        return StandardCharsets.UTF_8;
+      }
```
chingor13, line 526: Fallback code can be refactored if we want to add more defaults.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -512,8 +514,18 @@ public String parseAsString() throws IOException {
...
-    return mediaType == null || mediaType.getCharsetParameter() == null
-        ? StandardCharsets.ISO_8859_1
-        : mediaType.getCharsetParameter();
+    if (mediaType != null) {
+      // use specified charset parameter from content/type header if available
+      if (mediaType.getCharsetParameter() != null) {
+        return mediaType.getCharsetParameter();
+      }
+      // fallback to well-known charsets
+      if (mediaType.getType().equals("application") && mediaType.getSubType().equals("json")) {
```
elharo, line 523: possible null pointer exception here; invert

chingor13: Done.

elharo (approved):

PR #1315, opened 2021-03-11
fix: when disconnecting, close the underlying connection before the response InputStream

### LINKED ISSUE #1303: HttpResponse.disconnect() broken for ApacheHttpTransport
When using Apache as the underlying transport, issuing `response.getContent().close()` blocks waiting for the entire response to finish. If the source is the content from a GCS blob, this can potentially be many GB, which defeats the intention of `disconnect()`.

Calling `HttpResponse.disconnect()` before `.getContent().close()` falls on the problem that `disconnect()` [starts by calling `getContent().close()` before calling `LowLevelHttpHandler.disconnect()`](https://github.com/googleapis/google-http-java-client/blob/f17755cf5e8ccbf441131ebb13fe60028fb63850/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java#L435). The wrong order, compared to the [intended way for ApacheHttpClient](https://www.baeldung.com/httpclient-cancel-request).

The net effect is that the application-code get's blocked until all the blob-content have been read and discarded.

chingor13:
Adds a test to the `NetHttpTransport` and `ApacheHttpTransport` to make sure we don't wait until all the content is read when disconnecting the response.

Fixes #1303

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -351,9 +351,9 @@ public final class HttpResponse {
         try {
           // gzip encoding (wrap content with GZipInputStream)
           if (!returnRawInputStream && this.contentEncoding != null) {
-            String oontentencoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
-            if (CONTENT_ENCODING_GZIP.equals(oontentencoding)
-                || CONTENT_ENCODING_XGZIP.equals(oontentencoding)) {
+            String contentencoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
+            if (CONTENT_ENCODING_GZIP.equals(contentencoding)
+                || CONTENT_ENCODING_XGZIP.equals(contentencoding)) {
               // Wrap the original stream in a ConsumingInputStream before passing it to
               // GZIPInputStream. The GZIPInputStream leaves content unconsumed in the original
               // stream (it almost always leaves the last chunk unconsumed in chunked responses).
@@ -432,8 +432,10 @@ public final class HttpResponse {
    * @since 1.4
    */
   public void disconnect() throws IOException {
-    ignore();
+    // Close the connection before trying to close the InputStream content. If you are trying to
+    // disconnect, we shouldn't need to try to read any further content.
     response.disconnect();
+    ignore();
   }
 
   /**
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -419,9 +419,12 @@ public final class HttpResponse {
 
   /** Closes the content of the HTTP response from {@link #getContent()}, ignoring any content. */
   public void ignore() throws IOException {
-    InputStream content = getContent();
-    if (content != null) {
-      content.close();
+    if (this.response == null) {
+      return;
+    }
+    InputStream lowLevelResponseContent = this.response.getContent();
+    if (lowLevelResponseContent != null) {
+      lowLevelResponseContent.close();
     }
   }
 
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -351,9 +351,9 @@ public final class HttpResponse {
         try {
           // gzip encoding (wrap content with GZipInputStream)
           if (!returnRawInputStream && this.contentEncoding != null) {
-            String contentencoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
-            if (CONTENT_ENCODING_GZIP.equals(contentencoding)
-                || CONTENT_ENCODING_XGZIP.equals(contentencoding)) {
+            String contentEncoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
+            if (CONTENT_ENCODING_GZIP.equals(contentEncoding)
+                || CONTENT_ENCODING_XGZIP.equals(contentEncoding)) {
               // Wrap the original stream in a ConsumingInputStream before passing it to
               // GZIPInputStream. The GZIPInputStream leaves content unconsumed in the original
               // stream (it almost always leaves the last chunk unconsumed in chunked responses).
--- a/google-http-client-apache-v2/src/test/java/com/google/api/client/http/apache/v2/ApacheHttpTransportTest.java
+++ b/google-http-client-apache-v2/src/test/java/com/google/api/client/http/apache/v2/ApacheHttpTransportTest.java
@@ -268,6 +268,8 @@ public class ApacheHttpTransportTest {
 
   @Test(timeout = 10_000L)
   public void testDisconnectShouldNotWaitToReadResponse() throws IOException {
+    // This handler waits for 100s before returning writing content. The test should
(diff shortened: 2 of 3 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -351,9 +351,9 @@ public InputStream getContent() throws IOException {
         try {
           // gzip encoding (wrap content with GZipInputStream)
           if (!returnRawInputStream && this.contentEncoding != null) {
-            String oontentencoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
-            if (CONTENT_ENCODING_GZIP.equals(oontentencoding)
-                || CONTENT_ENCODING_XGZIP.equals(oontentencoding)) {
+            String contentencoding = this.contentEncoding.trim().toLowerCase(Locale.ENGLISH);
```
elharo, line 354: contentEncoding

elharo: also please triple check that the field isn't read somewhere below without `this`

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -432,8 +432,10 @@ public void ignore() throws IOException {
    * @since 1.4
    */
   public void disconnect() throws IOException {
-    ignore();
+    // Close the connection before trying to close the InputStream content. If you are trying to
+    // disconnect, we shouldn't need to try to read any further content.
     response.disconnect();
+    ignore();
```
elharo, line 438: Why do you need to call this at all? I'd think disconnecting alone is enough.

chingor13, line 438: I'm not sold this is the best solution -- the tests to pass however.

An alternative is to mark that this response is "disconnecting" and in `getContent()` don't read from the underlying response if we're "disconnecting".

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
chingor13: It is not. Will rename to `contentEncoding`

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
chingor13: I'm not sure of the original intention for sure, but it looks like it's supposed to close the response's `InputStream` so you can't read from it anymore - a way of letting google-http-client cleaning up the connections and their resources. This class was designed before `AutoClosable` was available (and possibly even before `Closable` was available).

Neenu1995 (approved):

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
elharo: The more I look at this, the more off this seems. I think what we should do is:

```
      InputStream lowLevelResponseContent = this.response.getContent();
      lowLevelResponseContent.close();
``
`ignore()` actually reads and decodes the bytes remaining on the stream. That could be quite a long-running operation. The point of disconnecting is to throw those away and not waste time reading them.

If you see otherwise, let's discuss.

**Review by elharo — CHANGES_REQUESTED** [review]

File: google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java (hunk 2 above)
Comment (chingor13) [reply google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java:438]: That seems reasonable, I'll try it out.

Comment (chingor13) [reply google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java:441]: I modified `ignore()` to directly close the `LowLevelHttpResponse`'s `InputStream` rather than doing the extra reading logic in `getContent()`.

I still do need to swap the `disconnect()` and `ignore()` lines as the `ApacheHttpResponse` will still try to read the content from its input stream unless it's disconnected first.

**Review by elharo — APPROVED** [review]

**Comment by suztomo** [comment]
@chingor13 I need help in upgrading this google-http-client dependency for Beam. With this PR 1315, google-http-client does not seem to read the content of responses with error status codes. Is this intended behavior when you made this pull request?

Background: I'm trying to upgrade the Beam's dependencies (https://github.com/apache/beam/pull/14527) and was updating assertions. To check the reason of the behavior change, I found this PR 1315 in google-http-client. Most of them are trivial (changing the number of invocation of `getStatusCode`, `getContent` and, `getContentType`), but I think certain error handling logic (throwing exceptions and logging messages to users) cannot work without reading the content of HTTP responses. Example assertion I'm stuck: https://github.com/apache/beam/pull/14527/files/7fe131c6256774134fc0a8503628c7add97b9984#r613669141

```
    when(response.getStatusCode()).thenReturn(403).thenReturn(200);
    // Google-http-client 1.39.1 and later does not read content stream of failed response
    when(response.getContent())
        .thenReturn(toStream(errorWithReasonAndStatus("rateLimitExceeded", 403)))
        .thenReturn(toStream(testTable));
...

    BigQueryServicesImpl.DatasetServiceImpl services =
        new BigQueryServicesImpl.DatasetServiceImpl(
            bigquery, null, PipelineOptionsFactory.create());
    Table ret =
        services.tryCreateTable(
            testTable, new RetryBoundedBackOff(3, BackOff.ZERO_BACKOFF), Sleeper.DEFAULT);
...
    // This logging does not work after upgrading google-http-client to 1.39.2 (from 1.38.1)
    expectedLogs.verifyInfo(
        "Quota limit reached when creating table project:dataset.table, "
            + "retrying up to 5.0 minutes");
```

The `tryCreateTable` method relies on `ApiErrorExtractor` to log errors appropriatey.

https://github.com/apache/beam/blob/4b7b74673b647c8d964b4877a8d66d47096acce4/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BigQueryServicesImpl.java#L626

```
        } catch (IOException e) {
          ApiErrorExtractor extractor = new ApiErrorExtractor();
          if (extractor.itemAlreadyExists(e)) {
            // The table already exists, nothing to return.
            return null;
          } else if (extractor.rateLimited(e)) {
```


I appreciate if you can help me to update the test case for the latest google-http-client.

**Comment by elharo** [comment]
http client is supposed to close the stream on a deliberate client invoked disconnect. not necessarily on an http error response. Is beam calling disconnect somewhere?

**Comment by elharo** [comment]
The problem might be this code in HttpRequest, in oarticular the call to disconnect when throwExceptionOnExecuteError is true.

```
    // throw an exception if unsuccessful response
    if (throwExceptionOnExecuteError && !response.isSuccessStatusCode()) {
      try {
        throw new HttpResponseException(response);
      } finally {
        response.disconnect();
      }
    }
```

However I can't yet prove this with a test.

elharo:
Are you trying to read the content from an exception?

elharo:
Might want to look into `extractor.rateLimited(e)` to see what it's doing. It might be incomplete or making invalid assumptions.

suztomo:
@chingor13 @elharo I now think the problem seems to lie in the mock setup in Beam. At least it has assumption that `lowLevelHttpRequest.getStatusCode()` is called only once per response. (Thank you for the idea of FakeServer!) I'll update you.

ChristianCiach:
This PR is the cause of issue https://github.com/GoogleContainerTools/jib/issues/3409. I don't know if this PR is broken by itself or if Jib somehow uses the google-http-java-client in the wrong way. I just want to bring this to your attention.

PR #1346, opened 2021-04-15
Test case that fails due to NullPointerException

suztomo:
Investigating the reason for https://github.com/apache/beam/pull/14527#discussion_r613980011


elharo (commented):
The mocking makes it hard to figure out whether there's a bug in the model code or just a bad mock setup. See https://github.com/googleapis/google-http-java-client/pull/1345 for a test that uses a real HTTP server instead of a mock.

elharo (commented):
Possibly BatchUnparsedResponse needs to handle empty input better

suztomo:
I agree that this test case is not clear which one is the problem. I'll try to simplify this further.

google-http-client-jackson2/src/test/java/com/google/api/client/json/jackson2/BatchTest.java
```diff
@@ -0,0 +1,199 @@
...
+
+    batch.execute();
+    assertNotNull(capturedGoogleJsonError[0]);
+
+    // From {"error":{"code":404}}
+    assertEquals(404, capturedGoogleJsonError[0].getCode());
+  }
+
+  @Test
+  public void testErrorContentRead_NoRetry() throws IOException {
```
suztomo, line 148: This succeeds. Maybe problem in retry?

elharo (commented):
I don't think this test demonstrates a bug or an issue with the http client. The NullPointerException here occurs in google-api-java-client. The code there looks very suspicious to me.

If there is an issue here it should be able to be demonstrated with classes already in this repository.

google-http-client-jackson2/src/test/java/com/google/api/client/json/jackson2/BatchTestWithFakeServer.java
```diff
@@ -0,0 +1,183 @@
...
+              byte[] response = content.getBytes(StandardCharsets.UTF_8);
+              responseHeaders.set("Content-Type", "multipart/mixed; boundary=" + contentBoundary);
+              httpExchange.sendResponseHeaders(200, response.length);
+              try (OutputStream out = httpExchange.getResponseBody()) {
+                out.write(response);
+              }
+            }
+          }
+        };
+    try (FakeServer server = new FakeServer(handler)) {
```
suztomo, line 105: With the fake server, this test succeeds. This implies that the mock setup in Beam is not enough (at least for this test case).

suztomo:
The FakeServer in #1345 worked. Thank you. Now I'm feeling something is lacking in the mock setup.

> If there is an issue here it should be able to be demonstrated with classes already in this repository

Right. I haven't been able to isolate the problem in a class yet.

suztomo:
Next step: investigate why mock fails and fake server succeeds.

PR #1411, opened 2021-07-12
Fix: Use BufferedInputStream to inspect HttpResponse error

### LINKED ISSUE #1410: Make a way to read HttpResponse content in HttpUnsuccessfulResponseHandler
Google API uses the `403 HttpStatus` for a variety of errors . To inspect the original error source you need to read `HttpResponse` content. But there is no way to do it in `HttpUnsuccessfulResponseHandler` and trigger `ExponentialBackOff`.

If you read `HttpResponse.getContent()` and decide not to process retry, the regular exception builder will cause IOException.

Here is the similar issue: #1403.

I suggest making `HttpResponse.getContent()` buffered in order to restore it's original state using `mark()/reset()`.

klya:
Fixes #1410

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -372,12 +372,13 @@ public final class HttpResponse {
                 new LoggingInputStream(
                     lowLevelResponseContent, logger, Level.CONFIG, contentLoggingLimit);
           }
-          if (!returnRawInputStream) {
+          if (returnRawInputStream) {
+            content = lowLevelResponseContent;
+          } else {
             // wrap the content with BufferedInputStream to support
             // mark()/reset() while error checking in error handlers
-            lowLevelResponseContent = new BufferedInputStream(lowLevelResponseContent);
+            content = new BufferedInputStream(lowLevelResponseContent);
           }
-          content = lowLevelResponseContent;
           contentProcessed = true;
         } catch (EOFException e) {
           // this may happen for example on a HEAD request since there no actual response data read
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -18,6 +18,7 @@ import com.google.api.client.util.IOUtils;
 import com.google.api.client.util.LoggingInputStream;
 import com.google.api.client.util.Preconditions;
 import com.google.api.client.util.StringUtils;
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -371,6 +372,11 @@ public InputStream getContent() throws IOException {
                 new LoggingInputStream(
                     lowLevelResponseContent, logger, Level.CONFIG, contentLoggingLimit);
           }
+          if (!returnRawInputStream) {
```
elharo, line 375: marginally cleaner to do a if-else

```
if (returnRawInputStream) {
  content = lowLevelResponseContent;
} else {
  content = new BufferedInputStream(lowLevelResponseContent);
}
```

This voids reassigning a local variable which can be confusing and error prone in the long term.

elharo (commented):
I'm not sure this is the best solution for the underlying issue of reading error responses. The whole mark/reset thing is a bit of a kludge. The error response should be more directly accessible in the API.

However this is a small, safe fix that provides a workaround until we can improve the API for this important case.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
klya: I chose this way just because all previous code in this method used this way. `lowLevelResponseContent` reassigned two times in preceding code.

klya:
> I'm not sure this is the best solution for the underlying issue of reading error responses. The whole mark/reset thing is a bit of a kludge. The error response should be more directly accessible in the API.

Yep. This way pretty awkward. I see alternatives like this:
* Fetching content from `HttpResponse` and passing it to the `HttpUnsuccessfulResponseHandler`. Nearly impossible due to braking changes in the `HttpUnsuccessfulResponseHandler` interface.
* Allowing access to the `HttpResponse.content` private field. But this is a violation of encapsulation.
* Make another kind of a handler that can wrap `HttpResponse.getContent()` or the entire `HttpResponse`. It looks rather complicated to implement and may cause some side-effects.

Can't find a better solution.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
elharo: Probably worth cleaning this up in a separate PR some time, but for now please don't introduce more reassignments

klya: done

elharo (approved):

PR #1423, opened 2021-08-05
fix: default charset to UTF-8 for text/csv if not specified

### LINKED ISSUE #1421: Default character set for text/csv to UTF-8 when absent in the response header.
Some servers don't return charset in the Content-Type HTTP header response. 

[Current implementation](https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java#L538) defaults to ISO_8859_1 in the absence of charset. 

ISO_8859_1 does not encode unicode characters correctly.

https://www.iana.org/assignments/media-types/text/csv and https://lists.w3.org/Archives/Public/public-csv-wg/2014Oct/0114.html 
indicate that in the absence of charset UTF-8 should be the assumed default. 

#### Environment details

1. General
2. OS type and version: All 
3. Java version: 11 
4. google-http-client version(s): latest 

#### External references such as API reference guides
https://www.iana.org/assignments/media-types/text/csv 
https://lists.w3.org/Archives/Public/public-csv-wg/2014Oct/0114.html

rohitvvv:
Some servers don't return the charset. This causes german
characters to be encoded incorrectly, since ISO_8859_1 does not
work very well in such cases defaulting to UTF-8 if its missing.

https://www.iana.org/assignments/media-types/text/csv

- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)

Fixes #1421  ☕️

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -534,6 +534,10 @@ public final class HttpResponse {
         // https://tools.ietf.org/html/rfc4627 - JSON must be encoded with UTF-8
         return StandardCharsets.UTF_8;
       }
+       // fallback to well-kown charset for text/csv
+      if ("text".equals(mediaType.getType()) && "csv".equals(mediaType.getSubType())) {
+        return StandardCharsets.UTF_8;
+      }
     }
     return StandardCharsets.ISO_8859_1;
   }
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
@@ -534,8 +534,9 @@ public final class HttpResponse {
         // https://tools.ietf.org/html/rfc4627 - JSON must be encoded with UTF-8
         return StandardCharsets.UTF_8;
       }
-       // fallback to well-kown charset for text/csv
+      // fallback to well-kown charset for text/csv
       if ("text".equals(mediaType.getType()) && "csv".equals(mediaType.getSubType())) {
+        // https://www.iana.org/assignments/media-types/text/csv - CSV must be encoded with UTF-8
         return StandardCharsets.UTF_8;
       }
     }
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -534,6 +534,10 @@ public Charset getContentCharset() {
         // https://tools.ietf.org/html/rfc4627 - JSON must be encoded with UTF-8
         return StandardCharsets.UTF_8;
       }
+       // fallback to well-kown charset for text/csv
```
chingor13, line 537: nit:
```suggestion
      // fallback to well-kown charset for text/csv
```

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
```diff
@@ -534,6 +534,10 @@ public Charset getContentCharset() {
...
+      if ("text".equals(mediaType.getType()) && "csv".equals(mediaType.getSubType())) {
+        return StandardCharsets.UTF_8;
```
chingor13, line 539: nit: can we add a comment with the link to the spec in case someone browsing the code wonders why this is here (without having to do a git blame and track down this PR through history).

chingor13 (changes requested):
LGTM, with 2 super-minor nits.

chingor13 (commented):
Also, please run `mvn com.coveo:fmt-maven-plugin:format` to fix the format/linter.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
rohitvvv: Done.

google-http-client/src/main/java/com/google/api/client/http/HttpResponse.java
rohitvvv: Done

rohitvvv:
Thank you for the review. I have updated the code.

chingor13 (approved):

PR #1486, opened 2021-10-18
GenericUrl: improve the error message for malformed URLs

alb-i986:
Let's say by mistake we pass a malformed URL to the constructor of `GenericUrl`.

```java
new GenericUrl("http://example.com:8080http://example.com:8080");
```

Previously the line above used to throw the following exception:

```
java.lang.IllegalArgumentException: java.net.MalformedURLException: Error at index 4 in: "8080http:"
```

Not very helpful.

Whereas with this patch:

```java
java.lang.IllegalArgumentException: Malformed URL: "http://example.com:8080http://example.com:8080/"
```

The exception thrown now clearly shows what was the faulty URL.
The user can therefore quickly figure out which URL needs to be fixed.

A lesson I learned today :)

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/GenericUrl.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/GenericUrl.java
@@ -676,7 +676,7 @@ public class GenericUrl extends GenericData {
     try {
       return new URL(encodedUrl);
     } catch (MalformedURLException e) {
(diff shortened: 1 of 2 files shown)


meltsufin (commented):
There might be a risk of leaking sensitive data from the URL. So, I'm leaning towards the JDK deciding what the message should be.

PR #1494, opened 2021-11-01
GH-1375 Improve performance of NetHttpRequest write timeouts by provding a possibility to use custom threadpool 

### LINKED ISSUE #1375: Improve performance of NetHttpRequest timeouts by reusing threads
If a timeout is specified for NetHttpRequest, a singlethreadedexecutor is created.
https://github.com/googleapis/google-http-java-client/blob/master/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpRequest.java#L184

This creates a new thread for every request which showed up as a 30% cpu increase issuing requests when writing to BigQuery from a Cloud Dataflow pipeline.

It seems that this cost could be reduced by using a cachedThreadPool, or by allowing the caller to inject an executorService to be used.

dzikoysk:
As described in #1375, creating a new thread per execution call pretty much kills the performance of application if it's oriented around high write throughput with configured write timeouts. 

The proposed solution allows user to provide custom thread executor, so it's now user's responsibility to handle this kind of scenarios and because it doesn't introduce any kind of strange internal cached threadpool, it's also easier to maintain in the future.

<details>
  <summary>Policy</summary>

- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)

</details>

Fixes #1375 ☕️


dzikoysk:
@googlebot I signed it!

dzikoysk:
Any thoughts on this PR? 🤔 Not sure if I didn't fulfill the requirements of PR, because CI hangs or it's a change you don't want to have in the codebase in general. I'm asking, because it's over a half year and I still need to use self-compiled & distributed artifact with this fix as the current impl is pretty unusable in high throughput environment with configured timeouts.

PR #1505, opened 2021-11-18
feat: add AttemptCount to HttpResponseException

TimurSadykov:
This adds number of requests attempted to httpRequestException. This makes sense to know how many attempts were made since HttpRequest supports automated retries. We want to use it to improve user experience of Auth library, that leverages HttpRequest retry functionality. 

For more details on the feature: go/auth-correct-retry

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
@@ -125,11 +125,11 @@ public class HttpResponseException extends IOException {
     return content;
   }
 
-  /** 
+  /**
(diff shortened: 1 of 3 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpResponseException.java
```diff
@@ -121,6 +125,15 @@ public final String getContent() {
     return content;
   }
 
+  /**
+   * Returns the attempt count
+   *
+   * @since 1.41
```
TimurSadykov, line 131: not sure if i need to add this manually

lesv (approved):

PR #1605, opened 2022-03-16
ci: not requiring lint until Owlbot migration finishes

suztomo:
@Neenu1995 The check came back as "Required". I think we need to edit this file.


suztomo:
I hope Owlbot doesn't override this file. (Maybe safe because Owlbot post processor passed)

PR #1716, opened 2022-09-01
fix: add `@CanIgnoreReturnValue` to avoid errorprone errors.

### LINKED ISSUE #1710: `NetHttpTransport.trustCertificates` ignores result of `SslUtils.initSslContext`
https://github.com/googleapis/google-http-java-client/blob/main/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java#L317

In the monorepo, this fails to compile.

```
google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java:317: error: [CheckReturnValue] The result of `initSslContext(...)` must be used
      SslUtils.initSslContext(
                             ^
  If you really don't want to use the result, then assign it to a variable: `var unused = ...`.
  
  If callers of `initSslContext(...)` shouldn't be required to use its result, then annotate it with `@CanIgnoreReturnValue`.
```

This is yet another case where if error prone was used and enforced in the external repo, we would not have import issues.

Does this code simply rely on the side effects on the passed in `sslContext`? OTOH, why return an `SSLContext` that was passed in as a variable? Chaining? Does `initSslContext` guarantee that the returned context is always the passed in context?

Capstan:
In the monorepo, this blocks the build from succeeding.

Fixes #1710.

### DIFF
--- a/google-http-client/pom.xml
+++ b/google-http-client/pom.xml
@@ -136,6 +136,10 @@
       <artifactId>jsr305</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.google.errorprone</groupId>
+      <artifactId>error_prone_annotations</artifactId>
+    </dependency>
+    <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
--- a/google-http-client/src/main/java/com/google/api/client/util/SslUtils.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/SslUtils.java
@@ -14,6 +14,7 @@
 
 package com.google.api.client.util;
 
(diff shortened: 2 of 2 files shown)


suztomo:
"ci / lint" has been failing and it's not required check. Don't worry about it. https://github.com/googleapis/google-http-java-client/pull/1691#issuecomment-1227485935

Capstan:
> "ci / lint" has been failing and it's not required check. Don't worry about it. [#1691 (comment)](https://github.com/googleapis/google-http-java-client/pull/1691#issuecomment-1227485935)

Sent separate pr #1717 to do lint cleanup.

suztomo (approved):

suztomo:
@blakeli0 (This week's release manager) Can you release this fix?

PR #1508, opened 2021-11-23
chore: troubleshooting lint problem

suztomo:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/google-http-java-client/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #<issue_number_goes_here> ☕️

PR #1717, opened 2022-09-01
chore: fix lint

Capstan:
This fixes the one error in the non-blocking linter CI results.


suztomo:
There was some mixed tooling (OwlBot) that prevented us from using the latest google-java-format.

https://github.com/googleapis/google-http-java-client/pull/1508

suztomo:
Trying "owlbot:run" label.

suztomo:
OwlBot reverted the format change because it uses an old version of google-java-format:

Screen Shot 2022-08-31 at 10 38 57 PM

Capstan:
Boo. I suppose I should just drop this then.

PR #1819, opened 2023-02-21
feat: GsonFactory to have read leniency option

suztomo:
Fixes b/269515918, where the prefix `)]}'\n` for Cross-site script inclusion (XSSI) didn't work for GsonFactory.

Users will call `GsonFactory.getDefaultInstance().withReadLeniency()` when they want this library to process JSON input that may have the prefix.

- For this pull request, it controlls only the leniency when reading input JSON. This design `withReadLeniency()` gives flexibility for us to choose writeLeniency in future if necessary. (It's possible that read leniency is true and write leniency is false.)
- Because this creates a copy of the instance, it does not affect existing usage of GsonFactory. (No global state in GsonFactory)

### DIFF
--- a/google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
+++ b/google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
@@ -98,7 +98,7 @@ public class GsonFactory extends JsonFactory {
     return readLeniency;
   }
 
-  private GsonFactory createCopy() {
+  protected GsonFactory createCopy() {
     GsonFactory copy = new GsonFactory();
     copy.readLeniency = this.readLeniency;
     return copy;
--- a/google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
+++ b/google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
@@ -52,7 +52,8 @@ public class GsonFactory extends JsonFactory {
     return InstanceHolder.INSTANCE;
   }
 
-  private boolean readLeniency = false;
+  /** Controls the behavior of leniency in reading JSON value */
+  protected boolean readLeniency = false;
 
   /** Holder for the result of {@link #getDefaultInstance()}. */
   @Beta
@@ -104,7 +105,12 @@ public class GsonFactory extends JsonFactory {
     return copy;
   }
 
-  /** Returns a copy of GsonFactory instance which is lenient when reading JSON value. */
+  /**
+   * Returns a copy of GsonFactory instance which is lenient when reading JSON value.
+   *
+   * <p>Subclasses should not call this method. Set {@code readLeniency} field to {@code true}
+   * instead.
+   */
   public GsonFactory withReadLeniency() {
     GsonFactory copy = newInstance(this);
     copy.readLeniency = true;
--- a/google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
+++ b/google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
@@ -62,7 +62,7 @@ public class GsonFactory extends JsonFactory {
   }
 
   // Keeping the default, non-arg constructor for backward compatibility. Users should use
-  // getDefaultInstance() or builder()
+  // getDefaultInstance() or builder() instead.
   public GsonFactory() {}
 
   private GsonFactory(Builder builder) {
@@ -113,7 +113,7 @@ public class GsonFactory extends JsonFactory {
   }
 
   /** Builder for GsonFactory. */
-  public static class Builder {
+  public static final class Builder {
     // Do not directly call this constructor
     private Builder() {}
 
--- a/google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
+++ b/google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
@@ -52,6 +52,8 @@ public class GsonFactory extends JsonFactory {
     return InstanceHolder.INSTANCE;
   }
 
+  private boolean readLeniency = false;
+
   /** Holder for the result of {@link #getDefaultInstance()}. */
   @Beta
   static class InstanceHolder {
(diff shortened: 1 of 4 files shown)


suztomo:
lint (non required) fails:

```
Error:  Found 1 non-complying files, failing build
Error:  To fix formatting errors, run "mvn com.coveo:fmt-maven-plugin:format"
Error:  Non complying file: /home/runner/work/google-http-java-client/google-http-java-client/google-http-client/src/test/java/com/google/api/client/http/HttpRequestTracingTest.java
```

This is irrelevant to this pull request.

google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
```diff
@@ -90,4 +92,22 @@ public JsonGenerator createJsonGenerator(OutputStream out, Charset enc) {
   public JsonGenerator createJsonGenerator(Writer writer) {
     return new GsonGenerator(this, new JsonWriter(writer));
   }
+
+  /** Returns true if it is lenient to input JSON value. */
+  boolean getReadLeniency() {
+    return readLeniency;
+  }
+
+  private GsonFactory createCopy() {
```
burkedavison, line 101: Consider `protected` since this class is not `final`.

suztomo: That's nice idea to let subclasses to define how to create a copy. Updating.

google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
```diff
@@ -90,4 +92,22 @@ public JsonGenerator createJsonGenerator(OutputStream out, Charset enc) {
   public JsonGenerator createJsonGenerator(Writer writer) {
     return new GsonGenerator(this, new JsonWriter(writer));
   }
+
+  /** Returns true if it is lenient to input JSON value. */
+  boolean getReadLeniency() {
+    return readLeniency;
+  }
+
+  private static GsonFactory newInstance(GsonFactory gsonFactory) {
```
suztomo, line 101: As per Effective Java's clone section, creating a factory method. I don't think a subclass needs to access this method.

google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
```diff
@@ -90,4 +92,22 @@ public JsonGenerator createJsonGenerator(OutputStream out, Charset enc) {
...
+    GsonFactory copy = new GsonFactory();
+    copy.readLeniency = gsonFactory.readLeniency;
+    return copy;
+  }
+
+  /** Returns a copy of GsonFactory instance which is lenient when reading JSON value. */
+  public GsonFactory withReadLeniency() {
+    GsonFactory copy = newInstance(this);
```
burkedavison, line 109: If we're supporting the subclassing use-case, I don't see how this works. Is the intent that subclasses are required to override this function? (If so, we're introducing a backwardly incompatible change to the subclass contract.)

As an example:
```java
// Logs all method invocations.
public LoggingGsonFactory extends GsonFactory { /* ... */ }

// In some 'main'-type class of the application...
public static GsonFactory createGsonFactory() { return new LoggingGsonFactory(); }

// Then elsewhere...
private GsonFactory gsonFactory = createGsonFactory().withReadLeniency(); 
// Oops! No more logging.
```

suztomo: Good point. Then subclasses needs to override newInstance. I'm making newInstance protected.

suztomo: But now it's static.

burkedavison: Consider avoiding the copy problem altogether.
```java
private boolean isLenient;
public GsonFactory() { this(false); }
public GsonFactory(boolean isLenient) { this.isLenient = isLenient; }
```

suztomo: The boolean flag in the constructor is not good. It's ambiguous, especially when we want to introduce leniency in writing JSON.

suztomo: Added `   * <p>Subclasses should not call this method. Set {@code readLeniency} field to {@code true}
   * instead.`.

suztomo (commented):
@burkedavison PTAL.

google-http-client-gson/src/main/java/com/google/api/client/json/gson/GsonFactory.java
burkedavison: "Subclasses should not call this method" is fine if you know you're dealing with a subclass. The point of polymorphism is that you shouldn't care.

There's other ways to take in a setting besides using a boolean if you feel it goes against best practices.

```java
public enum Leniency {
  LENIENT, STRICT
}
private final Leniency leniency;
public GsonFactory() { this(Leniency.STRICT); }
public GsonFactory(Leniency leniency) { this.leniency = Objects.requireNonNull(leniency); }
```
or a Builder, or a Settings object parameter.

suztomo: As talked, we may want to add more parameters (such as writeLeniency) in future and having different contstructors for different parameter is not good.

I'm bringing Builder class.

google-http-client-gson/src/test/java/com/google/api/client/json/gson/GsonFactoryTest.java
```diff
@@ -94,4 +100,30 @@ public final void testGetByteValue() throws IOException {
       assertNotNull(ex.getMessage());
     }
   }
+
+  public final void testReaderLeniency_lenient() throws IOException {
+    JsonObjectParser parser =
+        new JsonObjectParser(GsonFactory.builder().setReadLeniency(true).build());
```
suztomo, line 106: Memo: This is how to create GsonFactory with readLeniency true.

suztomo (commented):
@burkedavison PTAL.

burkedavison (approved):

google-http-client-gson/src/test/java/com/google/api/client/json/gson/GsonParserTest.java
```diff
@@ -13,13 +13,30 @@
...
+import java.nio.charset.StandardCharsets;
 
 public class GsonParserTest extends AbstractJsonParserTest {
 
   @Override
   protected JsonFactory newJsonFactory() {
     return new GsonFactory();
   }
+
+  public void testParse_leniency() throws IOException {
```
suztomo, line 33: This wasn't supposed to come here.

Issue #1830, opened 2023-03-10
JSON deserialization ignores case of setters


cushon:
The following example shows that json deserialization ignores the case of the setters, and deserializes two different fields that differ only in case to the same setter.

I stepped through in a debugger, and I think the issue is this logic that setups up a `FieldInfo` for deserialization ignores case:

https://github.com/googleapis/google-http-java-client/blob/84216c5cfe2f1600464d34661a208cf165e11b9b/google-http-client/src/main/java/com/google/api/client/util/FieldInfo.java#L140-L141

That logic was added as part of https://github.com/googleapis/google-http-java-client/commit/ff934795716ae3083d1edc8078bb3843e559e59a

---

Repro

With `com.google.http-client:google-http-client-gson:1.28.0`

```
Serialized: {"passCode":"pass1","passcode":"pass2"}
{"passCode":"pass2"}
```

With `com.google.http-client:google-http-client-gson:1.27.0`

```
Serialized: {"passCode":"pass1","passcode":"pass2"}
{"passCode":"pass1","passcode":"pass2"}
```

```java
import com.google.api.client.json.GenericJson;
import com.google.api.client.json.gson.GsonFactory;
import com.google.api.client.util.Key;

import java.io.StringReader;
import java.io.StringWriter;

public class Repro {

  public static final class Data extends GenericJson {
    @Key String passcode;
    @Key String passCode;

    public String getPasscode() {
      return passcode;
    }

    public Data setPasscode(String passcode) {
      this.passcode = passcode;
      return this;
    }

    public String getPassCode() {
      return passCode;
    }

    public Data setPassCode(String passCode) {
      this.passCode = passCode;
      return this;
    }
  }

  public static void main(String[] args) throws Exception {
    String serialized;
    {
      Data c = new Data().setPassCode("pass1").setPasscode("pass2");
      StringWriter writer = new StringWriter();
      new GsonFactory().createJsonGenerator(writer).serialize(c);
      serialized = writer.toString();
    }
    System.out.println("Serialized: " + serialized);

    Data parsedData =
        new GsonFactory()
            .createJsonObjectParser()
            .parseAndClose(new StringReader(serialized), Data.class);

    System.out.println(parsedData);
  }
}
```

PR #1831, opened 2023-03-11
fix: JSON deserialization setter case priority

meltsufin:
Fixes the problem of JSON deserialization ignoring the case of setters. Now case-sensitive setter method matches are considered first.

Fixes: #1830.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/FieldInfo.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/FieldInfo.java
@@ -134,12 +134,21 @@ public class FieldInfo {
   }
 
   /** Creates list of setter methods for a field only in declaring class. */
-  private Method[] settersMethodForField(Field field) {
+  private Method[] settersMethodForField(final Field field) {
     List<Method> methods = new ArrayList<>();
+    String fieldSetter = "set" + Ascii.toUpperCase(field.getName().substring(0, 1));
+    if (field.getName().length() > 1) {
+      fieldSetter += field.getName().substring(1);
+    }
     for (Method method : field.getDeclaringClass().getDeclaredMethods()) {
-      if (Ascii.toLowerCase(method.getName()).equals("set" + Ascii.toLowerCase(field.getName()))
-          && method.getParameterTypes().length == 1) {
-        methods.add(method);
+      if (method.getParameterTypes().length == 1) {
+        // add case-sensitive matches first in the list
+        if (method.getName().equals(fieldSetter)) {
+          methods.add(0, method);
+        } else if (Ascii.toLowerCase(method.getName())
+            .equals(Ascii.toLowerCase(fieldSetter))) {
+          methods.add(method);
+        }
       }
     }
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/util/FieldInfo.java
```diff
@@ -134,12 +134,20 @@ public static FieldInfo of(Field field) {
...
+    }
     for (Method method : field.getDeclaringClass().getDeclaredMethods()) {
-      if (Ascii.toLowerCase(method.getName()).equals("set" + Ascii.toLowerCase(field.getName()))
-          && method.getParameterTypes().length == 1) {
-        methods.add(method);
+      if (method.getParameterTypes().length == 1) {
+        // add case-sensitive matches first in the list
+        if (method.getName().equals(fieldSetter)) {
+          methods.add(0, method);
+        } else if (Ascii.toLowerCase(method.getName()).equals(Ascii.toLowerCase(fieldSetter))) {
```
burkedavison, line 148: I wish I understood the reason behind this lowercase comparison better. It feels like it was an arbitrary implementation choice while fixing an issue.

I don't think we can change it after all this time, though.

Original PR: https://github.com/googleapis/google-http-java-client/pull/538

burkedavison (approved):

google-http-client/src/main/java/com/google/api/client/util/FieldInfo.java
```diff
@@ -134,12 +134,20 @@ public static FieldInfo of(Field field) {
   }
 
   /** Creates list of setter methods for a field only in declaring class. */
-  private Method[] settersMethodForField(Field field) {
+  private Method[] settersMethodForField(final Field field) {
     List<Method> methods = new ArrayList<>();
+    String fieldSetter = "set" + Ascii.toUpperCase(field.getName().substring(0, 1));
```
blakeli0, line 139: This approach works for this issue, is it possible to have a scenario that the field name is `utcTime`, and the setter is `setUTCTime`? We may have to look at how GenericJson is implemented to confirm this. So I think it would be safer to split the `method.getName()` by `set` and compare the name with `field.getName()`.

meltsufin: The default setter for `utcTime` would be `setUtcTime()`. So, I think this logic would work. The default setter would take precedence over `setUTCTime()`.

suztomo:
I didn’t know this case priority behavior. If not done yet, can you write down the desired behavior about the setter case priority in the Javadoc? Include how it helps the users of this HTTP client library.

If the current implementation was not doing the job in the spec, then this PR is a bug fix.

meltsufin:
> I didn’t know this case priority behavior. If not done yet, can you write down the desired behavior about the setter case priority in the Javadoc? Include how it helps the users of this HTTP client library.
> 
> If the current implementation was not doing the job in the spec, then this PR is a bug fix.

It's a bugfix. Internal ref: b/267807198.

PR #1901, opened 2023-11-22
feat: add isShutdown in HttpTransport

### LINKED ISSUE #1850: Investigate and Implement `shutdown()` and `isShutdown()` for NetHttpTransport
From: https://github.com/googleapis/sdk-platform-java/pull/1677#discussion_r1198241529

It would be nice to be able to have a method to determine if a transport has been properly shutdown. Investigate the different clients and how to handle logic for determining if it's been shutdown or not.

JoeWang1127:
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)

Fixes #1850 ☕️

If you write sample code, please follow the [samples format](
https://github.com/GoogleCloudPlatform/java-docs-samples/blob/main/SAMPLE_FORMAT.md).

### DIFF
--- a/google-http-client-appengine/src/main/java/com/google/api/client/extensions/appengine/http/UrlFetchTransport.java
+++ b/google-http-client-appengine/src/main/java/com/google/api/client/extensions/appengine/http/UrlFetchTransport.java
@@ -32,13 +32,12 @@ import java.util.Arrays;
  *
  * <p>URL Fetch is only available on Google App Engine (not on any other Java environment), and is
  * the underlying HTTP transport used for App Engine. Their implementation of {@link
- * HttpURLConnection} is simply an abstraction layer on top of URL Fetch. By implementing
- * transport that directly uses URL Fetch, we can optimize the behavior slightly, and can
- * potentially take advantage of features in URL Fetch that are not available in {@link
- * HttpURLConnection}. Furthermore, there is currently a serious bug in how HTTP headers are
- * processed in the App Engine implementation of {@link HttpURLConnection}, which we are able to
- * avoid using this implementation. Therefore, this is the recommended transport to use on App
- * Engine.
+ * HttpURLConnection} is simply an abstraction layer on top of URL Fetch. By implementing transport
+ * that directly uses URL Fetch, we can optimize the behavior slightly, and can potentially take
+ * advantage of features in URL Fetch that are not available in {@link HttpURLConnection}.
+ * Furthermore, there is currently a serious bug in how HTTP headers are processed in the App Engine
+ * implementation of {@link HttpURLConnection}, which we are able to avoid using this
+ * implementation. Therefore, this is the recommended transport to use on App Engine.
  *
  * @since 1.10
  * @author Yaniv Inbar
--- a/google-http-client-appengine/src/main/java/com/google/api/client/extensions/appengine/http/UrlFetchTransport.java
+++ b/google-http-client-appengine/src/main/java/com/google/api/client/extensions/appengine/http/UrlFetchTransport.java
@@ -32,12 +32,13 @@ import java.util.Arrays;
  *
  * <p>URL Fetch is only available on Google App Engine (not on any other Java environment), and is
  * the underlying HTTP transport used for App Engine. Their implementation of {@link
- * HttpURLConnection} is simply an abstraction layer on top of URL Fetch. By implementing transport
- * that directly uses URL Fetch, we can optimize the behavior slightly, and can potentially take
- * advantage of features in URL Fetch that are not available in {@link HttpURLConnection}.
- * Furthermore, there is currently a serious bug in how HTTP headers are processed in the App Engine
- * implementation of {@link HttpURLConnection}, which we are able to avoid using this
- * implementation. Therefore, this is the recommended transport to use on App Engine.
+ * HttpURLConnection} is simply an abstraction layer on top of URL Fetch. By implementing a
+ * transport that directly uses URL Fetch, we can optimize the behavior slightly, and can
+ * potentially take advantage of features in URL Fetch that are not available in {@link
+ * HttpURLConnection}. Furthermore, there is currently a serious bug in how HTTP headers are
+ * processed in the App Engine implementation of {@link HttpURLConnection}, which we are able to
+ * avoid using this implementation. Therefore, this is the recommended transport to use on App
+ * Engine.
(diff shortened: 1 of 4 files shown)


google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
```diff
@@ -158,4 +158,13 @@ public boolean isMtls() {
    * @since 1.4
    */
   public void shutdown() throws IOException {}
+
+  /**
+   * Default implementation of whether the http transport is shutdown or not.
+   *
+   * @since 1.4
+   */
+  public boolean isShutdown() {
```
blakeli0, line 167: It would be better to return false in the default implementation and override it to true in the child classes, similar to how `isMtls()` was implemented above. Or throw `UnsupportedOperationException()`, similar to what we did for [GAPIC stubs](https://github.com/googleapis/google-cloud-java/blob/main/java-asset/google-cloud-asset/src/main/java/com/google/cloud/asset/v1/stub/AssetServiceStub.java#L95).

google-http-client-appengine/src/main/java/com/google/api/client/extensions/appengine/http/UrlFetchTransport.java
```diff
@@ -32,13 +32,12 @@
...
  * <p>URL Fetch is only available on Google App Engine (not on any other Java environment), and is
  * the underlying HTTP transport used for App Engine. Their implementation of {@link
- * HttpURLConnection} is simply an abstraction layer on top of URL Fetch. By implementing a
- * transport that directly uses URL Fetch, we can optimize the behavior slightly, and can
- * potentially take advantage of features in URL Fetch that are not available in {@link
- * HttpURLConnection}. Furthermore, there is currently a serious bug in how HTTP headers are
- * processed in the App Engine implementation of {@link HttpURLConnection}, which we are able to
- * avoid using this implementation. Therefore, this is the recommended transport to use on App
- * Engine.
+ * HttpURLConnection} is simply an abstraction layer on top of URL Fetch. By implementing transport
```
blakeli0, line 35: Can you highlight the change in this paragraph? Or is it just format changes?

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
```diff
@@ -158,4 +158,13 @@ public boolean isMtls() {
    * @since 1.4
    */
   public void shutdown() throws IOException {}
+
+  /**
+   * Default implementation of whether the http transport is shutdown or not.
+   *
+   * @since 1.4
```
blakeli0, line 165: This version should be latest.

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
```diff
@@ -158,4 +158,13 @@ public boolean isMtls() {
    * @since 1.4
    */
   public void shutdown() throws IOException {}
+
+  /**
+   * Default implementation of whether the http transport is shutdown or not.
```
blakeli0, line 163: The Javadoc for methods in abstract classes should focus on the expected behavior, as the subclasses will be referring to it if an overriding is needed, e.g. [ExecutorService.isShutdown()](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/util/concurrent/ExecutorService.html#isShutdown()). We can mention the default implementation but should not be the focus.

google-http-client-appengine/src/main/java/com/google/api/client/extensions/appengine/http/UrlFetchTransport.java
JoeWang1127: This is doc format change.

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
JoeWang1127: Done, thanks for the suggestion.

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
JoeWang1127: Done, changed to 1.44.0 (the next release).

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
JoeWang1127: Changed to `false`.

google-http-client-appengine/src/main/java/com/google/api/client/extensions/appengine/http/UrlFetchTransport.java
JoeWang1127: On a second thought, comment change on other files are unrelated to this PR, so I restore all comment changes.

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010 Google Inc.
+ * Copyright (c) 2023 Google Inc.
```
blakeli0, line 2: We don't want to update the year on new changes, see go/copyright#the-year

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
blakeli0: We would need to override it to `true` in NetHttpTransport if the default is `false`.
However, given it more thought, I think it makes more sense to return `true` as default, as you initially proposed. Because if a class extends this abstract class and don't override anything, `shutdown()` and `isShutdown()` should have consistent behavior, meaning that an no-op `shutdown()` should be accompanied with `isShutdown()` equal to `true`.

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
JoeWang1127: done.

google-http-client/src/main/java/com/google/api/client/http/HttpTransport.java
JoeWang1127: Sure, I changed it back to `true`.

blakeli0 (approved):

Issue #1933, opened 2024-03-21
External documentation is outdated - latest published version is 1.43.2


alicejli:
The last version of this library that's been published to c.g.c. is v1.43.2 (https://cloud.google.com/java/docs/reference/google-http-client/latest/overview.html). I double-checked the Cloud RAD bucket, and confirmed that v1.43.2 is the latest tarball that's been uploaded to the bucket.

This means there's something wrong with the release process for this library such that the tarball doesn't get generated/uploaded to the Cloud RAD bucket. I suspect the migration of the release jobs to Kokoro from Github likely broke this.

PR #1934, opened 2024-03-21
chore: remove obsolete kokoro configs and update DocFX profile

alicejli:
Partial fix for https://github.com/googleapis/google-http-java-client/issues/1933 (the other part of the fix will be in g3)

This removes the kokoro release configs as that was migrated to g3. This also updates the DocFX profile such that it javadoc generation will work. This repo doesn't inherit from java-shared-config AFAIK.


alicejli:
We can't add Java 21 to the series of tests due to compilation error:
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project google-http-client: Compilation failure: Compilation failure: 
Error:  Source option 7 is no longer supported. Use 8 or later.
Error:  Target option 7 is no longer supported. Use 8 or later.
Error:  -> [Help 1]
```

That's the reasoning behind the custom test that compiles with 8 and then tests with 21:  https://github.com/googleapis/google-http-java-client/blob/main/.github/workflows/ci.yaml#L90

burkedavison (approved):

suztomo:
Do we use docFX for this repository?

suztomo:
Never mind. Found it https://cloud.google.com/java/docs/reference/google-http-client/latest/overview.

PR #1941, opened 2024-05-15
fix: Base64 decoding to discard newline characters

suztomo:
Using b/274482271#comment3 by @eamonnmcmanus

- Added test cases for new line characters.
- This PR keeps the `trim()` method that was also added for the compatibility.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/Base64.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/Base64.java
@@ -26,10 +26,13 @@ import com.google.common.io.BaseEncoding.DecodingException;
  */
 @Deprecated
 public class Base64 {
-  // Special decoders that discards the new line character so that the behavior matches what
-  // we had with Apache Commmons Codec's decodeBase64.
-  private static final BaseEncoding BASE64 = BaseEncoding.base64().withSeparator("\n", 64);
-  private static final BaseEncoding BASE64URL = BaseEncoding.base64Url().withSeparator("\n", 64);
+  // Guava's Base64 (https://datatracker.ietf.org/doc/html/rfc4648#section-4) decoders. When
+  // decoding, they discard the new line character so that the behavior matches what we had with
+  // Apache Commons Codec's decodeBase64. When encoding, they would insert a new line character
+  // every 64 (the 2nd argument of withSeparator method) characters. They are not used for encoding.
+  private static final BaseEncoding BASE64_DECODER = BaseEncoding.base64().withSeparator("\n", 64);
+  private static final BaseEncoding BASE64URL_DECODER =
+      BaseEncoding.base64Url().withSeparator("\n", 64);
 
   /**
    * Encodes binary data using the base64 algorithm but does not chunk the output.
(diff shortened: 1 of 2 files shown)


google-http-client/src/main/java/com/google/api/client/util/Base64.java
```diff
@@ -26,6 +26,10 @@
  */
 @Deprecated
 public class Base64 {
+  // Special decoders that discards the new line character so that the behavior matches what
+  // we had with Apache Commmons Codec's decodeBase64.
```
eamonnmcmanus, line 30: I think it would be worth commenting on what the 64 means. (It doesn't mean very much, since it's only used for encoding and we don't do that.)

eamonnmcmanus (changes requested):
LGTM, with just one very small thing.

google-http-client/src/main/java/com/google/api/client/util/Base64.java
suztomo: Added explanation of the 2nd parameter of the withSeparator method.

google-http-client/src/main/java/com/google/api/client/util/Base64.java
```diff
@@ -26,6 +26,13 @@
  */
 @Deprecated
 public class Base64 {
+  // Guava's Base64 (https://datatracker.ietf.org/doc/html/rfc4648#section-4) decoders. When
+  // decoding, they discard the new line character so that the behavior matches what we had with
+  // Apache Commons Codec's decodeBase64. When encoding, they would insert a new line character
+  // every 64 (the 2nd argument of withSeparator method) characters. They are not used for encoding.
```
eamonnmcmanus, line 32: The last word should be "decoding". :-)

suztomo: They (the two variables) are not used for encoding.

I'll rewrite this to make it clear.

suztomo: Done

google-http-client/src/test/java/com/google/api/client/util/Base64Test.java
```diff
@@ -62,4 +64,62 @@ public void test_encodeBase64URLSafe_withNull_shouldReturnNull() {
...
+    //        org.apache.commons.codec.binary.Base64.decodeBase64(encodedString),
+    //        StandardCharsets.UTF_8));
+
+    // This is our implementation. Before the
+    // https://github.com/googleapis/google-http-java-client/pull/1941/, it was throwing
+    // IllegalArgumentException("Unrecognized character: 0xa").
+    assertEquals("ab", new String(Base64.decodeBase64(encodedString), StandardCharsets.UTF_8));
+  }
+
+  public void test_decodeBase64__plus_and_newline_characters() {
```
diegomarquezp, line 115: May be slightly overkill, but we could also include a test for the `/` charater as per b/274482271#comment4

suztomo: I can add that.

suztomo: Added

eamonnmcmanus (approved):
Sorry for misunderstanding in my previous comment.

PR #2014, opened 2024-09-11
chore(apache-v5): confirm behavior of connect timeout

diegomarquezp:
This is to confirm the connection timeout doesn't file like it does in https://github.com/firebase/firebase-admin-java/pull/979


diegomarquezp:
Tested in https://github.com/firebase/firebase-admin-java/pull/979/commits/11aab962a9e6022ef8cba4d9874c0f3e2333e103 (thanks @jonathanedey!)

PR #2069, opened 2025-02-06
feat: Update and adapt to GraalVM for JDK 23

diegomarquezp:
This PR updates the parent project of this repository "native-image-shared-config" to 1.14.0, which introduces GraalVM for JDK 23. Several adaptations were necessary. All changes listed here:

 * Update native-image-shared-config to v1.14.0
 * Adapt all modules to JUnit4
   * Annotate all test classes with `RunWith(JUnit4.class)`
   * Exclude/remove/replace all JUnit 3.x methods and classes
   * Add "test time" native image initialization flags for JUnit classes (e.g. `--initialize-at-build-time=org.junit.RunWith`. These classes don't go in production code. The `google-http-client-jackson2` module had most of `jackson.core` module initialized at build time **only in the test image** for simplicity.
 * Initialize all static fields in a `static` block
 * Avoid "cryptic" `<clinit>` messages using `try-catch` blocks that log class initialization errors. This provides extra information when creating native images.
 * The test module `google-http-client-test` had a few tweaks related to upgrading to JUnit4, which made some CLIRR-ignored changes necessary. They will not affect production code since this is a production-level test-only library.
 * Tweaked reflect configuration of a few modules.


google-http-client-appengine/pom.xml
```diff
@@ -72,7 +72,6 @@
     <dependency>
       <groupId>com.google.appengine</groupId>
       <artifactId>appengine-api-1.0-sdk</artifactId>
-      <scope>provided</scope>
```
ldetmer, line 75: i'm hesitant to remove this, as this makes our overall jar larger.  Could we instead add this as another class to the test only -initialize-at-build-time ?

google-http-client-gson/pom.xml
```diff
@@ -72,5 +72,10 @@
       <groupId>com.google.code.gson</groupId>
       <artifactId>gson</artifactId>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
```
ldetmer, line 76: same here, I'm curious why this wasn't needed for JDK < 23

google-http-client-jackson2/src/main/java/com/google/api/client/json/jackson2/JacksonFactory.java
```diff
@@ -60,7 +60,11 @@ public static JacksonFactory getDefaultInstance() {
 
   /** Holder for the result of {@link #getDefaultInstance()}. */
   static class InstanceHolder {
-    static final JacksonFactory INSTANCE = new JacksonFactory();
+    static final JacksonFactory INSTANCE;
+
+    static {
```
ldetmer, line 65: is this static block necessary so that we don't have to include this file at build time?

google-http-client-test/clirr-ignored-differences.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ldetmer, line 1: really quick, is there a way to ignore test classes for backwards binary compatibility?  they should be out of scope

google-http-client-test/pom.xml
```diff
@@ -59,7 +59,6 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
-      <scope>provided</scope>
```
ldetmer, line 62: same here, if we can instead include this for native image I would prefer that

ldetmer (commented):
Thanks Diego, just a couple of questions/clarifications

google-http-client-appengine/pom.xml
diegomarquezp: The -initialize-at-build-time flag has no effect when using this. However I believe you can specify native-specific dependencies. I'll try to find a way around.

google-http-client-appengine/src/main/resources/META-INF/native-image/com.google.http-client/google-http-client-appengine/reflect-config.json
```diff
@@ -0,0 +1,47 @@
+[
+    {
+      "name": "com.google.apphosting.api.DatastorePb$DeleteRequest",
```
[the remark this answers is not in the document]
diegomarquezp: Looks like Graal SDK 22 introduced a new class initialization strategy called `strict image heap`, which is [enforced by default](https://www.graalvm.org/release-notes/JDK_22/).

Since this is a behavior change affecting which classes are initialized, I'm leaning towards this new strategy being the cause.

google-http-client-test/pom.xml
diegomarquezp: Will try

google-http-client-test/clirr-ignored-differences.xml
diegomarquezp: This one is a bit strange. It's a production module with JUnit-annotated classes in it's `src/main` subfolders, which causes CLIRR to treat it as any other production module.
Do you mean that we could skip CLIRR for this module? (it only contains test classes)

google-http-client-test/src/main/java/com/google/api/client/test/json/AbstractJsonFactoryTest.java
```diff
@@ -335,22 +372,27 @@ public static class NumberTypesAsString {
...
 
-  static final String NUMBER_TYPES_AS_STRING =
-      "{\"bigDecimalValue\":\"1.0\",\"bigIntegerValue\":\"1\",\"byteObjValue\":\"1\","
-          + "\"byteValue\":\"1\",\"doubleObjValue\":\"1.0\",\"doubleValue\":\"1.0\","
-          + "\"floatObjValue\":\"1.0\",\"floatValue\":\"1.0\",\"intObjValue\":\"1\","
-          + "\"intValue\":\"1\",\"longListValue\":[\"1\"],\"longMapValue\":{\"a\":\"1\"},"
-          + "\"longObjValue\":\"1\",\"longValue\":\"1\","
-          + "\"shortObjValue\":\"1\","
-          + "\"shortValue\":\"1\",\"yetAnotherBigDecimalValue\":\"1\"}";
+  static {
```
mpeddada1, line 378: I can see how this may be needed for more complex setup or when initializing a class (which can sometimes result in `NoClassDefFoundError`)  but do we still need this static block if we're only initializing a string constant?

google-http-client-appengine/pom.xml
diegomarquezp: Note: there was a bug issued last year specifically talking about this. https://github.com/graalvm/native-build-tools/issues/562

google-http-client-jackson2/src/main/java/com/google/api/client/json/jackson2/JacksonFactory.java
diegomarquezp: I reverted simple static initializations. I only kept a couple that had more complex logic + a trycatch.

google-http-client-test/pom.xml
diegomarquezp: This was always fixed via the `-Pnative-deps` profile. I restored this.

google-http-client-test/src/main/java/com/google/api/client/test/json/AbstractJsonFactoryTest.java
diegomarquezp: Indeed not, I think it was a wrong conclusion from solving a NoClassDefFoundError while preparing this PR that I extrapolated to the whole repo. I reverted this so only actual complex logic is moved into a static block with trycatch (only 2 cases)

mpeddada1: Thank you.

mpeddada1 (approved):
LGTM

google-http-client-test/clirr-ignored-differences.xml
ldetmer: Since this is a public module, I think technically its possible that someone could be using it to run their own tests. If that's the case as long as we make sure the breaking changes show up in release notes I'm okay with ignoring them here.  But lets leave the module included in breaking change detector so we can always add to release notes.

ldetmer (approved):
thanks for the updates.  Please just ensure the test breaking changes make it into release notes.

google-http-client-test/clirr-ignored-differences.xml
diegomarquezp: Ok, I edited the release notes in the release PR explaining the breaking change in this module. https://github.com/googleapis/google-http-java-client/pull/2061

PR #2151, opened 2026-02-20
fix: add containsKey override for GenericData class

logachev:
When running JFR profiling for Google BigQuery JDBC driver, I saw large amount of samples points to `Map.containsKey` function. 

image

When I tried local changes to BQ Java SDK to avoid `containsKey` call and call `get()` instead & validate if result is null or not, calls to `get` function, they don't generate as many samples. 

Due to `containsKey` override missing, it falls back to the default `Map.containsKey` implementation which seems to be more expensive in the case of `GenericData` class. I didn't run sampler with these changes, but attaching results when `containsKey` was replaced with `get()` in the SDK.

image

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/util/ClassInfo.java
+++ b/google-http-client/src/main/java/com/google/api/client/util/ClassInfo.java
@@ -113,6 +113,23 @@ public final class ClassInfo {
   }
 
   /**
+   * Returns whether information for the given {@link FieldInfo#getName()}
+   * is available.
+   *
+   * @param name {@link FieldInfo#getName()} or {@code null}
+   * @return true if field info is available.
+   */
+  public boolean hasFieldInfo(String name) {
+    if (name != null) {
+      if (ignoreCase) {
+        name = name.toLowerCase(Locale.US);
+      }
+      name = name.intern();
+    }
+    return nameToFieldInfoMap.containsKey(name);
+  }
+  
+  /**
    * Returns the information for the given {@link FieldInfo#getName()}.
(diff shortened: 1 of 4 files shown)


diegomarquezp (approved):

google-http-client/src/main/java/com/google/api/client/util/ClassInfo.java
```diff
@@ -112,6 +112,22 @@ public final boolean getIgnoreCase() {
     return ignoreCase;
   }
 
+  /**
+   * Returns whether information for the given {@link FieldInfo#getName()} is available.
+   *
+   * @param name {@link FieldInfo#getName()} or {@code null}
+   * @return true if field info is available.
+   */
+  public boolean hasFieldInfo(String name) {
```
lqiu96, line 121: nit: Is this just for testing? Or do we need this for the new change in BQ?

If just for testing, can we make this package-private?

logachev: oops, good catch; I meant to use it in `GenericData`. Updated

lqiu96 (approved):

blakeli0:
Can we change the title from `chore` to `fix`?

cpovirk:
This appears to be a behavior change, one that produces a surprising number of failures in the Google monorepo and might well do the same elsewhere.

I don't know enough to know whether the behavior change is an improvement or not. Googlers can peek at bug 539478616, which reports on that and one other set of failures that results from recent changes to the project.

PR #2161, opened 2026-05-25
chore: POC for PQC support using Bouncy Castle

diegomarquezp:


google-http-client/src/main/java/com/google/api/client/util/SslUtils.java
```diff
@@ -191,4 +271,229 @@ public boolean verify(String arg0, SSLSession arg1) {
...
+      this.delegate = delegate;
+    }
+
+    @Override
+    public void setSSLParameters(javax.net.ssl.SSLParameters params) {
+      delegate.setSSLParameters(params);
+      Object objEngine = delegate;
+      if (objEngine instanceof org.bouncycastle.jsse.BCSSLEngine) {
+        org.bouncycastle.jsse.BCSSLEngine bcEngine = (org.bouncycastle.jsse.BCSSLEngine) objEngine;
+        org.bouncycastle.jsse.BCSSLParameters bcParams = bcEngine.getParameters();
```
diegomarquezp, line 289: If bouncy castle can achieve this with is own BCSSLParameters, then we can see if we can do something similar with Conscrypt.

PR #2167, opened 2026-06-29
feat(javanet): Introduce configurable SecurityProviders and SSLSocketConfigurator for NetHttpTransport 

lqiu96:
Adds support to configure the SSLSocket. This change allows users a method to enable Post Quantum Cryptography (PQC), but can be used for any other SSLSocket configuration. Allows a method for to configure a SecurityProvider and modify the the corresponding socket factory with the SecurityProvider.

There is no default SecurityProvider set in the http-client and Gax and Java-Core will bundle Conscrypt by default as pass it into the http-client. The Java SDK uses Conscrypt as the default SecurityProvider for PQC and will use the Conscrypt native APIs (e.g. Conscrypt.setNamedGroups) via the SslSocketConfigurator interface.

Changes: 
-  Adds SslSocketConfigurator: A generic callback interface invoked on newly created SSLSocket instances right before the TLS handshake starts.
-  Adds ConfigurableSSLSocketFactory: An SSLSocketFactory wrapper that intercepts all socket creation entrypoints (including bound socket creations) and applies the custom SslSocketConfigurator callback to them.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
@@ -28,12 +28,15 @@ import java.net.Proxy;
 import java.net.URL;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
+import java.security.Provider;
 import java.security.cert.CertificateFactory;
 import java.util.Arrays;
 import javax.net.ssl.HostnameVerifier;
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
 
 /**
  * Thread-safe HTTP low-level transport based on the {@code java.net} package.
@@ -84,7 +87,7 @@ public final class NetHttpTransport extends HttpTransport {
   private final ConnectionFactory connectionFactory;
 
   /** SSL socket factory or {@code null} for the default. */
-  private final SSLSocketFactory sslSocketFactory;
+  final SSLSocketFactory sslSocketFactory;
 
   /** Host name verifier or {@code null} for the default. */
   private final HostnameVerifier hostnameVerifier;
@@ -189,6 +192,12 @@ public final class NetHttpTransport extends HttpTransport {
     /** SSL socket factory or {@code null} for the default. */
     private SSLSocketFactory sslSocketFactory;
 
+    /** Security provider to use or {@code null} for default. */
+    private Provider securityProvider;
+
+    /** Custom SSLSocket configurator or {@code null} to disable callback configuration. */
+    SslSocketConfigurator sslSocketConfigurator;
+
     /** Host name verifier or {@code null} for the default. */
     private HostnameVerifier hostnameVerifier;
 
@@ -289,8 +298,9 @@ public final class NetHttpTransport extends HttpTransport {
      * @since 1.14
      */
     public Builder trustCertificates(KeyStore trustStore) throws GeneralSecurityException {
-      SSLContext sslContext = SslUtils.getTlsSslContext();
-      SslUtils.initSslContext(sslContext, trustStore, SslUtils.getPkixTrustManagerFactory());
+      SSLContext sslContext = SslUtils.getTlsSslContext(securityProvider);
+      SslUtils.initSslContext(
+          sslContext, trustStore, SslUtils.getPkixTrustManagerFactory(securityProvider));
       return setSslSocketFactory(sslContext.getSocketFactory());
     }
 
@@ -313,14 +323,14 @@ public final class NetHttpTransport extends HttpTransport {
       if (mtlsKeyStore != null && mtlsKeyStore.size() > 0) {
         this.isMtls = true;
       }
-      SSLContext sslContext = SslUtils.getTlsSslContext();
+      SSLContext sslContext = SslUtils.getTlsSslContext(securityProvider);
       SslUtils.initSslContext(
           sslContext,
           trustStore,
-          SslUtils.getPkixTrustManagerFactory(),
+          SslUtils.getPkixTrustManagerFactory(securityProvider),
           mtlsKeyStore,
           mtlsKeyStorePassword,
-          SslUtils.getDefaultKeyManagerFactory());
+          SslUtils.getDefaultKeyManagerFactory(securityProvider));
       return setSslSocketFactory(sslContext.getSocketFactory());
     }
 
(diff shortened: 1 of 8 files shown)


google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
```diff
@@ -362,14 +388,86 @@ public Builder setHostnameVerifier(HostnameVerifier hostnameVerifier) {
...
+     * Resolves and instantiates the {@link SSLContext} for "TLS", prioritizing the configured
+     * custom security provider, falling back to inserting and loading the Conscrypt provider, and
+     * defaulting to the JRE's standard SSLContext if Conscrypt is unavailable.
+     */
+    private SSLContext createSslContext() throws GeneralSecurityException {
+      // 1. If a custom security provider is configured, use it
+      if (securityProvider != null) {
+        return SSLContext.getInstance(TLS_ALGORITHM, securityProvider);
+      }
+      // 2. Default: Try Conscrypt (assumed to be available as part of SDK)
```
blakeli0: Can we always pass the Conscrypt provider in? So this library may not have to be dependent on Conscrypt?

google-http-client/src/main/java/com/google/api/client/http/javanet/PqcEnforcingSSLSocketFactory.java
```diff
@@ -0,0 +1,199 @@
...
+  private Socket configure(Socket socket) {
+    if (socket instanceof SSLSocket) {
+      SSLSocket sslSocket = (SSLSocket) socket;
+      try {
+        // 1. If it is a Conscrypt socket, use Conscrypt's direct JNI API first
+        if (Conscrypt.isConscrypt(sslSocket)) {
+          Conscrypt.setNamedGroups(sslSocket, groups);
+          return socket;
+        }
+        // 2. Otherwise (user-provided security provider or JDK fallback), try standard JSSE
```
blakeli0: Is this a case that we have to support? This is a generic http library and I don't think we have to try to enable PQC by default. As long as we expose an option and users (including GAPIC/Apiary libraries) can set the PQC algorithm then we should be good.

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
lqiu96: I think that is an option. Right now that's the intention of provided scope, library isn't dependent on Conscrypt and will only enable it by default if it's used via a library in the Java SDK

google-http-client/src/main/java/com/google/api/client/http/javanet/PqcEnforcingSSLSocketFactory.java
lqiu96: I'll double check the requirements. The earlier design mentioned that we will support PQC by default in the http-client. This functionality mirrors what gRPC-Java is doing which enables it by default (both being general purpose libraries)

google-http-client/src/main/java/com/google/api/client/http/javanet/JreNamedGroupsSslSocketConfigurator.java
blakeli0, line 1: This class does not seem to be used in this PR or gax. Is it for customers that use http-client as a standalone library and want PQC support?

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
```diff
@@ -362,14 +404,64 @@ public Builder setHostnameVerifier(HostnameVerifier hostnameVerifier) {
...
+        return (SSLSocketFactory) SSLSocketFactory.getDefault();
+      }
+    }
+
     /** Returns a new instance of {@link NetHttpTransport} based on the options. */
     public NetHttpTransport build() {
       if (System.getProperty(SHOULD_USE_PROXY_FLAG) != null) {
         setProxy(defaultProxy());
       }
+      SSLSocketFactory resolvedFactory = resolveSslSocketFactory();
```
blakeli0, line 485: The default value of sslSocketFactory was null but we create a default one now?

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
```diff
@@ -345,12 +355,44 @@ public SSLSocketFactory getSslSocketFactory() {
...
       this.sslSocketFactory = sslSocketFactory;
       return this;
     }
 
+    /**
+     * Sets the custom security provider or {@code null} to use default.
+     *
+     * @param securityProvider provider to use
+     */
+    public Builder setSecurityProvider(Provider securityProvider) {
```
blakeli0, line 389: Seems like both `securityProvider` and `sslSocketConfigurator` need to be set for PQC support. Is there any chance they are misconfigured? e.g. only one is set, both are set but they are not compatible etc.

google-http-client/src/main/java/com/google/api/client/http/javanet/JreNamedGroupsSslSocketConfigurator.java
lqiu96: > Is it for customers that use http-client as a standalone library and want PQC support?

Yep, but also for these possible cases:
1. JDK 20+ and do not want to use Conscrypt, BouncyCastle, or any alternative and prefer the native JDK
2. JDK 20+, but use a version of Conscrypt or BouncyCastle that does not have a native API to set the named groups so they use the JDK's SSLParameters way.

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
lqiu96: I think it may be a bit confusing so I'll update the javadocs to make it clearer. They work in tandem but aren't necessarily both required to enable PQC support.

1. SecurityProvider allows users to swap between something like Conscrypt, BouncyCastle, or alternative
2. sslSocketConfigurator allows users to configure how they want modify the SSLSocket (does't necessarily need to modify the NamedGroups for PQC, can theoretically be anything)

If a user has JDK 27+, then they do not need to use either for PQC support as it comes in by default by the JDK (they can choose to override if they want).
JDK 20+ would require that SSLParameters.setNamedGroups() be called to pass in the algos if they do not configure it via the library's native APIs
For JDK 8-20, then Conscrypt or a different SecurityProvider would be needed to use the native APIs.

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
lqiu96: Updated. Should now respect the default `sslSocketFactory` for the JDK unless explicitly configured.

google-http-client/src/main/java/com/google/api/client/http/javanet/JreNamedGroupsSslSocketConfigurator.java
blakeli0: Discussed offline that this is a nice-to-have helper but not required.

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
```diff
@@ -362,14 +429,65 @@ public Builder setHostnameVerifier(HostnameVerifier hostnameVerifier) {
...
+      try {
+        SSLContext sslContext = SslUtils.getTlsSslContext(securityProvider);
+        TrustManager[] trustManagers = null;
+        if (securityProvider != null) {
+          try {
+            TrustManagerFactory tmf = SslUtils.getDefaultTrustManagerFactory(securityProvider);
+            tmf.init((KeyStore) null);
+            trustManagers = tmf.getTrustManagers();
+          } catch (Exception e) {
+            // Ignore and fall back to default
```
blakeli0: Why this case can be ignored? Shall we log it at least? 
Also `trustManagers` will be null in this. case, would it cause issues when `sslContext.init` is called?

google-http-client/src/main/java/com/google/api/client/util/SslUtils.java
```diff
@@ -54,6 +55,70 @@ public static SSLContext getTlsSslContext() throws NoSuchAlgorithmException {
...
+   * <p>If a custom provider (e.g., Conscrypt) is configured, the context must be loaded from it to
+   * enable the provider's specific TLS parameters and curve groups (such as PQC).
+   *
+   * @param provider the security provider, or {@code null} to use the default JRE provider
+   * @since 2.1.2
+   */
+  public static SSLContext getTlsSslContext(Provider provider) throws NoSuchAlgorithmException {
+    return provider != null
+        ? SSLContext.getInstance("TLS", provider)
+        : SSLContext.getInstance("TLS");
```
blakeli0: Delegate to `getTlsSslContext()` instead? Similar comments for all the following methods.

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
```diff
@@ -362,14 +429,65 @@ public Builder setHostnameVerifier(HostnameVerifier hostnameVerifier) {
...
+    SSLSocketFactory createDefaultSslSocketFactory() {
+      try {
+        SSLContext sslContext = SslUtils.getTlsSslContext(securityProvider);
+        TrustManager[] trustManagers = null;
+        if (securityProvider != null) {
+          try {
+            TrustManagerFactory tmf = SslUtils.getDefaultTrustManagerFactory(securityProvider);
+            tmf.init((KeyStore) null);
+            trustManagers = tmf.getTrustManagers();
+          } catch (Exception e) {
```
blakeli0: If we know what exactly the checked exceptions are, it is better to use the exact exception.

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
```diff
@@ -362,14 +429,65 @@ public Builder setHostnameVerifier(HostnameVerifier hostnameVerifier) {
...
+            // Ignore and fall back to default
+          }
+        }
+        sslContext.init(null, trustManagers, null);
+        return sslContext.getSocketFactory();
+      } catch (Exception e) {
```
blakeli0: Same comment as https://github.com/googleapis/google-http-java-client/pull/2167/changes#r3605325675. Unless it is meant to be a "catch all and fall back to default" case.

google-http-client/src/main/java/com/google/api/client/util/SslUtils.java
blakeli0, line 1: Can we add unit tests for these new methods?

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
lqiu96: Hmm, updating this so we just propagate errors instead. If we can't initialize the custom socket configuration, it may be better to just fail fast instead of silently change to JDK (I will assume there is a reason they require a specific securityprovideR)

google-http-client/src/main/java/com/google/api/client/util/SslUtils.java
lqiu96: done

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
lqiu96: i'm going to remove catching exceptions alltogether

google-http-client/src/main/java/com/google/api/client/util/SslUtils.java
lqiu96: added some additional test cases

google-http-client/src/main/java/com/google/api/client/http/javanet/NetHttpTransport.java
lqiu96: Updated. On second thought, I don't think we should fall back to the default. I think fail-fast would make more sense here

blakeli0 (approved):

PR #2168, opened 2026-06-30
fix: declare commons-codec as direct dependency in google-http-client

lqiu96:
This ensures that consumers of google-http-client (like google-auth-library) resolve commons-codec to the intended version (1.14) instead of falling back to the transitive default from httpclient (1.11).

### Verification
Below is the comparison of `mvn dependency:tree` on `google-auth-library-oauth2-http` before and after this change, demonstrating that `commons-codec` is now correctly resolved to `1.14` transitively.
#### Before (with google-http-client v2.1.1)
`commons-codec` resolved to `1.11` via `httpclient`:
```text
[INFO] +- com.google.http-client:google-http-client:jar:2.1.1:compile
[INFO] |  +- org.apache.httpcomponents:httpclient:jar:4.5.14:compile
[INFO] |  |  \- commons-codec:commons-codec:jar:1.11:compile
```

#### After
```
[INFO] +- com.google.http-client:google-http-client:jar:2.1.2-SNAPSHOT:compile
[INFO] |  +- org.apache.httpcomponents:httpclient:jar:4.5.14:compile
[INFO] |  +- commons-codec:commons-codec:jar:1.14:compile
```

### DIFF
--- a/google-http-client/pom.xml
+++ b/google-http-client/pom.xml
@@ -25,6 +25,7 @@
             <ignoredUnusedDeclaredDependencies>
               <ignoredUnusedDeclaredDependency>io.opencensus:opencensus-impl</ignoredUnusedDeclaredDependency>
               <ignoredUnusedDeclaredDependency>io.grpc:grpc-context</ignoredUnusedDeclaredDependency>
+              <ignoredUnusedDeclaredDependency>commons-codec:commons-codec</ignoredUnusedDeclaredDependency>
             </ignoredUnusedDeclaredDependencies>
           </configuration>
         </plugin>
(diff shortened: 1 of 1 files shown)


blakeli0:
> This ensures that consumers of google-http-client (like google-auth-library) resolve commons-codec to the intended version (1.14) instead of falling back to the transitive default from httpclient (1.11).

I think it should be the consumer's responsibility to resolve to a higher version? For GAPIC client libraries, it should be fine since [commons-codec is already managed](https://github.com/googleapis/google-cloud-java/blob/ece312631154a6c13dd5a5daedc26c6559993b0e/sdk-platform-java/java-shared-dependencies/third-party-dependencies/pom.xml#L146). So maybe we manage the version in auth as well? In the future though, we should extract it to a high level so both auth and third-party-dependencies share the same versions.

lqiu96:
> I think it should be the consumer's responsibility to resolve to a higher version?

Hmm, I would think this applies for direct dependencies that consumers need a newer version for or transitive deps pulled from repos where we can't change ourselves. It seems odd to have downstream libraries manage the versions for upstream transitive dependencies if the deps come from upstream repos we own. 

I guess we already manage deps for a few transitive deps in shared-deps for better or worse and I'll keep with the pattern. 

> So maybe we manage the version in auth as well

Yeah, I'll update in auth bom for now

PR #2169, opened 2026-07-01
security: strip cookies on cross-origin redirects and fix response connection leaks

insaf021:
Summary

Cross-Origin Cookie Removal on Redirects 

Problem: When `HttpRequest` follows a redirect to a different origin (different scheme, host, or effective port), sensitive `Cookie` headers could be forwarded to the redirect destination, potentially exposing session credentials.

Fix: Updated `handleRedirect()` to compare the redirect target's scheme, host, and effective port with the current request URL. If the redirect is cross-origin, the `Cookie` header is removed before following the redirect. Same-origin redirects continue to preserve cookies.

 Resource Cleanup on Response Construction Failure 

Problem: If `new HttpResponse(this, lowLevelHttpResponse)` throws an exception before the response is fully constructed, the underlying `LowLevelHttpResponse` may not be released, potentially leaving underlying HTTP resources open.

Fix: Updated the cleanup logic in the `finally` block of `execute()` to close the response content (when present) and invoke `disconnect()` on `lowLevelHttpResponse` when response construction fails, ensuring cleanup exceptions do not mask the original exception.

Tests Added 

Added `testHandleRedirect_crossOriginCookieRemoval` to verify that cookies are removed for cross-origin redirects while remaining intact for same-origin redirects, including cases involving implicit and explicit default ports. Added `testExecute_disconnectOnResponseConstructionFailure` to verify that `disconnect()` is invoked when response construction fails and that the original exception is preserved during cleanup.

 Verification  

All 45 tests in `HttpRequestTest` passed successfully. A full `mvn clean install` also completed successfully across all 15 modules with `BUILD SUCCESS`, confirming the changes behave as expected and do not introduce regressions.

### DIFF
--- a/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
+++ b/google-http-client/src/main/java/com/google/api/client/http/HttpRequest.java
@@ -29,6 +29,7 @@ import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.util.Properties;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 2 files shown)


insaf021:
any update?

PR #2170, opened 2026-07-04
security: mitigate cross-origin credential leaks, prevent CRLF header injection, and disable XML DTD processing

insaf021:
This PR implements security hardening improvements in the `google-http-java-client` library to strengthen credential handling across redirects, improve HTTP header validation, and harden XML parsing.

When handling redirects, the original request origin is recorded and compared against the redirected request. If a redirect targets a different origin (scheme, host, or port), sensitive headers such as `Authorization` and `Cookie` are removed after execution interceptors run, preventing credentials from being forwarded to cross-origin destinations while preserving existing behavior for same-origin redirects. Unit tests have been added to verify both same-origin and cross-origin redirect behavior.

This PR also adds validation in `HttpHeaders#addHeader` to reject header names and serialized header values containing carriage return (`\r`) or line feed (`\n`) characters by throwing an `IllegalArgumentException`. Corresponding tests verify that invalid headers are rejected while valid headers continue to serialize correctly.

Additionally, `Xml#createParser()` now disables DTD processing using `XmlPullParser.FEATURE_PROCESS_DOCDECL` where supported by the parser implementation, with a graceful fallback for implementations that do not support the feature. Tests have been added to verify that XML documents containing DTD declarations do not process external entities while maintaining normal XML parsing behavior.

The implementation has been validated by running the relevant unit tests. `HttpRequestTest`, `HttpHeadersTest`, and `XmlTest` all pass successfully, with no test failures or errors.

### DIFF
--- a/google-http-client-xml/src/main/java/com/google/api/client/xml/Xml.java
+++ b/google-http-client-xml/src/main/java/com/google/api/client/xml/Xml.java
@@ -87,7 +87,13 @@ public class Xml {
 
   /** Returns a new XML pull parser. */
   public static XmlPullParser createParser() throws XmlPullParserException {
-    return getParserFactory().newPullParser();
+    XmlPullParser parser = getParserFactory().newPullParser();
+    try {
+      parser.setFeature(XmlPullParser.FEATURE_PROCESS_DOCDECL, false);
+    } catch (XmlPullParserException e) {
+      // Ignore if the feature is not supported
+    }
+    return parser;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


elharo:
This sounds like multiple separate issues. I'm no longer at Google so I don;t know how likely external contributions are to be considered but in general:

1. File individual issues for each item.
2. Break the PR up into multiple PRs, each addressing one thing.

Grab bag PRs tend not to go anywhere.

insaf021:
Splitting this into focused, single-purpose PRs per your feedback:

- #2178 — fix response connection leak (LowLevelHttpResponse not disconnected)
- #2182 — strip Authorization/Cookie headers on cross-origin redirects
- #2183 — disable XML DTD processing to prevent XXE
- #2184 — validate headers to prevent CRLF injection

Each has its own issue and isolated diff. Closing this one in favor of those four.

